PR #9, opened 2021-05-20
Use map definition for key and value sizes instead of arguments

brancz:
Previously this was quite error prone and somewhat unnecessary, as one
had to keep sizes consistent in Go and C code, but we can actually read
the sizes from the map definition improving developer experience and
making the API less error prone.

Already verified these changes in my usage of libbpfgo.

@grantseltzer @yanivagman @eyakubovich

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -217,6 +217,7 @@ type Module struct {
 
 type BPFMap struct {
 	name   string
+	bpfMap *C.struct_bpf_map
 	fd     C.int
 	module *Module
 }
@@ -365,6 +366,7 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	}
 
 	return &BPFMap{
+		bpfMap: bpfMap,
 		name:   mapName,
 		fd:     C.bpf_map__fd(bpfMap),
 		module: m,
@@ -457,17 +459,21 @@ func GetUnsafePointer(data interface{}) (unsafe.Pointer, error) {
 	return dataPtr, nil
 }
 
-func (b *BPFMap) GetValue(key interface{}, valueSize int) ([]byte, error) {
+func (b *BPFMap) KeySize() int {
+	return int(C.bpf_map__key_size(b.bpfMap))
+}
+
+func (b *BPFMap) ValueSize() int {
+	return int(C.bpf_map__value_size(b.bpfMap))
+}
+
+func (b *BPFMap) GetValue(key interface{}) ([]byte, error) {
 	keyPtr, err := GetUnsafePointer(key)
 	if err != nil {
 		return nil, fmt.Errorf("failed to lookup value in map %s: unknown key type %T", b.name, key)
 	}
 
-	// Create pointer to value byte array
-	if valueSize == 0 {
-		return nil, fmt.Errorf("valueSize must be > 0")
-	}
-	value := make([]byte, valueSize)
+	value := make([]byte, b.ValueSize())
 	valuePtr := unsafe.Pointer(&value[0])
 
 	errC := C.bpf_map_lookup_elem(b.fd, keyPtr, valuePtr)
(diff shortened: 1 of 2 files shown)


libbpfgo.go
```diff
@@ -365,9 +366,10 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	}
 
 	return &BPFMap{
-		name:   mapName,
-		fd:     C.bpf_map__fd(bpfMap),
-		module: m,
+		definition: C.bpf_map__def(bpfMap),
```
grantseltzer, line 369: What do you think about exposing API for getting the map definition? We have one for getting max entries right now which is only part of the map definition.

brancz: That would be strictly additional to what's in this PR no? Happy to discuss what this should look like in an issue and do a follow-up.

grantseltzer: Correct, it could be a follow up PR, that likely wouldn't change anything in the code you changed here.

grantseltzer:
I opened a PR to add a self test here: https://github.com/aquasecurity/libbpfgo/pull/11. Could you take a look at that, and I'd merge it first, so you can rebase this PR on top of it more safely?

libbpfgo.go
grantseltzer: Actually, I take back my suggestion. I'm re-reading the [libbpf 1.0 roadmap](https://docs.google.com/document/d/1UyjTZuPFWiPFyKk1tV5an11_iaRuec6U-ZESZ54nNTY/edit#), and it states that using `bpf_map__def()` should be discouraged in favor of getter APIs for the individual fields of the bpf map definition. In this case we can just use [`bpf_map__value_size()`](https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L459) here.

I'm making issues for the individual map definition APIs now!

brancz:
#41 looks good! Will rebase this one on top of it once merged!

grantseltzer:
Merged #11, so feel free to rebase and please check my comment about using `bpf_map__value_size()` directly instead of `bpf_map__def()`. Great work on this!

brancz:
Sorry for taking so long to update this PR. It should be ready now and incorporates the use of `bpf_map__value_size` and `bpf_map__key_size`.

grantseltzer:
@brancz No need to apologize, your contribution is very much appreciated! 

It also looks great, it's just missing updating the iterator selftest for usage of the new API changes your introducing. If you're not too busy could you update it?

brancz:
Updated! (I'm having trouble running the selftest so it would be good to double-check that)

grantseltzer:
@brancz Thanks, it looks like it works. What trouble are you having running the selftests? I imagine it has to do with linking to a system libbpf which I fix with #19, but can you paste the error just to be sure?

brancz:
```
$ ./run.sh 
    [*] SUFFICIENT KERNEL VERSION
CC=gcc CGO_CFLAGS="-I /usr/include/bpf" CGO_LDFLAGS="/usr/lib64/libbpf.a" go build -ldflags '-extldflags "-static"' -o self 
# github.com/aquasecurity/libbpfgo/selftest/iterators
self.bpf.c:10:1: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘#pragma’
   10 | char LICENSE[] SEC("license") = "Dual BSD/GPL";
      | ^~~
self.bpf.c:10:1: error: expected identifier or ‘(’ before ‘#pragma’
self.bpf.c:17:1: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘#pragma’
   17 | } numbers SEC(".maps");
      | ^~~
self.bpf.c:17:1: error: expected identifier or ‘(’ before ‘#pragma’
make: *** [/home/brancz/src/github.com/aquasecurity/libbpfgo/selftest/iterators/Makefile:20: self] Error 2
    [*] MAKE FAILED
```

brancz:
(I checked out #19 and I'm getting the same error)

grantseltzer:
Hm, have you ever compiled Tracee? if so did you ever have issues? I remember having this issue once before but don't remember exactly what the root cause was. It had something to do with the packages I had installed. Like an outdated version of libbpf-devel. 

Either way, i'll an open an issue to further investigate and merge this since it seems unrelated.

Thanks again @brancz! Please feel free to let us know what features you and your team need in libbpfgo and we can change priorities around as need be.

grantseltzer (approved):
💯

brancz:
Compiling tracee works like a charm. We need to measure a bit, but I think batch operations are going to be next for us.

PR #20, opened 2021-06-02
Uprobe support

grantseltzer:
This allows users to use the libbpf API for uprobes. This gives you the ability to safely attach to symbols in a userspace binary. 

The API allows users to specify a pid, binary, and symbol name. We could instead have the API allow users to specify specific offsets (instead of symbols) and then make the symbol->offset resolution a public helper function. 

Before we merge this i'd like to verify how architecture specific this is. Right now symbol resolution is handled by parsing ELF binaries. We should think about if there's another object format we should support

### DIFF
--- a/helpers/elf.go
+++ b/helpers/elf.go
@@ -2,13 +2,13 @@ package helpers
 
 import (
 	"debug/elf"
+	"errors"
 	"fmt"
 )
 
-// SymbolToOffset attempts to resolve a 'symbol' name within a specific
-// 'section' of the binary found at 'path' to an offset. The offset
-// can be used for attaching a u(ret)probe
-func SymbolToOffset(path, section, symbol string) (uint32, error) {
+// SymbolToOffset attempts to resolve a 'symbol' name in the binary found at
+// 'path' to an offset. The offset can be used for attaching a u(ret)probe
+func SymbolToOffset(path, symbol string) (uint32, error) {
 
 	f, err := elf.Open(path)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


selftest/uprobe/test_program/test_program.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	"os"
+)
+
+//go:noinline
+func testFunction() int {
+	if os.Getenv("SELFTEST_UPROBE_DO_BRANCH") == "y" {
```
grantseltzer, line 9: I have this here to demonstrate attaching to offsets not at the top of a symbol, which could make the argument in favor of changing the API to take an offset instead of a symbol name.

yanivagman:
From my experience with uprobes, it is very usefull to be able to attach a uprobe to a specific address, and not only by symbol. I think that we should add another API function that will do that.
Elf support should be enough for most use cases, so I think there is no problem to have only that for now.

libbpfgo.go
```diff
@@ -750,6 +754,90 @@ func doAttachKprobe(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink, error
...
+	if err != nil {
+		return 0, fmt.Errorf("could not open elf file to resolve symbol offset: %v", err)
+	}
+
+	syms, err := f.Symbols()
+	if err != nil {
+		return 0, fmt.Errorf("could not open symbol section to resolve symbol offset: %v", err)
+	}
+
+	textSection := f.Section(".text")
```
yanivagman, line 802: I'm not sure that only looking into the .text section is enough.
Although usually this will work, there are cases where the text section is named differently or there is more than one text section (with different names)

selftest/uprobe/go.sum
```diff
@@ -0,0 +1,16 @@
+github.com/aquasecurity/libbpfgo v0.1.0 h1:14V7ISF45XVqwBre+7NCDByy77OuwFgzDjVR+0JltZ8=
+github.com/aquasecurity/libbpfgo v0.1.0/go.mod h1:/+clceXE103FaXvVTIY2HAkQjxNtkra4DRWvZYr2SKw=
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
simar7, line 4: are these needed? maybe you forgot a `go mod tidy`?

grantseltzer: I'm not sure why this is, but I did run a go mod tidy (which made changes) but this is still there. It's imported by testify

libbpfgo.go
grantseltzer: I changed the AttachU(ret)probe API to accept an offset. I moved the ResolveOffset helper function to the helpers package. That function takes a symbol name and section, so leave it in our users hands 👍

grantseltzer: Example in selftest

grantseltzer:
@yanivagman @simar7 could you take another look at this when you get a chance?

simar7:
> @yanivagman @simar7 could you take another look at this when you get a chance?

I'll try later today.

simar7 (approved):
lgtm

selftest/uprobe/go.sum
simar7: ah yeah, probably testify brought it in

yanivagman (approved):
We can merge this as is, but please see my comment about the new section argument.

helpers/elf.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"debug/elf"
+	"fmt"
+)
+
+// SymbolToOffset attempts to resolve a 'symbol' name within a specific
+// 'section' of the binary found at 'path' to an offset. The offset
+// can be used for attaching a u(ret)probe
+func SymbolToOffset(path, section, symbol string) (uint32, error) {
+
```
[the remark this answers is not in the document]
grantseltzer: Is it possible for colliding symbol names? If not I think searching for it would be a great idea.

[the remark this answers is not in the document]
yanivagman: I don't think that there can be colliding symbol names on the same ELF file. The linker would have given an error if that was the case.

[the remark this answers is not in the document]
grantseltzer: Just pushed these changes 👍

yanivagman (approved):

PR #23, opened 2021-06-03
Fix eyes on tux logo to be white instead of transparent

grantseltzer:
This way he doesn't look like a ghost with github dark mode


simar7 (approved):
tux lives matter

PR #24, opened 2021-06-13
add tc api

yanivagman:
This PR adds the low level tc api as described in https://github.com/libbpf/libbpf/commit/d71ff87a2dd7b92787719aab233767e9c74fbd48

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -196,6 +196,7 @@ import "C"
 import (
 	"fmt"
 	"path/filepath"
+	"net"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -1038,3 +1039,143 @@ func (pb *PerfBuffer) poll() error {
...
+	iface, err := net.InterfaceByName(ifaceName)
+	if err != nil {
+		return err
+	}
+	hook.hook.ifindex = C.int(iface.Index)
+
+	return nil
+}
+
+func (hook *TcHook) GetInterfaceIndex() int {
```
grantseltzer, line 1119: What would users need this for?

grantseltzer (commented):
Do I understand correctly that this allows users to attach bpf programs to tc actions, where tc actions are when tc filters are modified in some ways? 

I.e.) User adds a tc filter for slowing down egress on a particular interface. The bpf program would have access to information about that action.

libbpfgo.go
yanivagman: Not sure that this would be needed, but I thought it may be good to have. I don't mind removing it though.

yanivagman:
> Do I understand correctly that this allows users to attach bpf programs to tc actions, where tc actions are when tc filters are modified in some ways?
> 
> I.e.) User adds a tc filter for slowing down egress on a particular interface. The bpf program would have access to information about that action.

grantseltzer (approved):
LGTM!

PR #29, opened 2021-07-06
Fix error message in BPFMap.Update

derekparker:
When trying to obtain an unsafe pointer to the value type, upon failure
the code was incorrectly returning an error message which contained the
type of the key, not the value. This patch fixes it so that the correct
type is returned for the value in the error message.

Additionally this commit contains a small refactor which uses a type
switch rather than if-else statements for readability.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -195,8 +195,8 @@ import "C"
 
 import (
 	"fmt"
-	"path/filepath"
 	"net"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -442,21 +442,22 @@ func (b *BPFMap) GetMaxEntries() uint32 {
 
 func GetUnsafePointer(data interface{}) (unsafe.Pointer, error) {
 	var dataPtr unsafe.Pointer
-	if k, isType := data.(int8); isType {
+	switch k := data.(type) {
```
yanivagman, line 445: Nice!
I wasn't aware of this type switch option.
And now I'm asking myself - do we even need the switch-case?
All the cases do the same thing - `dataPtr = unsafe.Pointer(&k)`
maybe just fallthrough?

yanivagman: Ok, after reading about type switches I understand that we must do it this way so that k will have the correct type asserted.

yanivagman (approved):
LGTM

libbpfgo.go
derekparker: Yeah, this could actually be refactored further into something like:

```
switch k := data.(type) {
case <all ints>: ....
case []byte: ....
default: ....
```

However I'm curious why this function is so restrictive in the first place. I've actually spoken with @grantseltzer offline about this and plan on hacking on this further in order to allow more arbitrary types, if possible.

yanivagman: When I wrote this code, I tried to find a generic method to get a pointer to the concrete type so I can pass it to libbpf, but couldn't find any. So I just coded the types that were needed for tracee, which I guess are the most common ones.
I'll be happy to change this code to something more generic if there is a sple way to do that

derekparker: Ack, makes sense!

PR #28, opened 2021-07-01
Makefile: improvements

rafaeldtinoco:
Makefile: improvements for the entire tree

Currently you will find the following GNU Makefile rules:

| Makefile Rule &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Description &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; |
|--------------------------|-----------------------------------|
| all                      | builds libbpfgo (dynamic)         |
| clean                    | cleans entire tree                |
| selftest                 | builds all selftests (static)     |
| selftest-run             | runs all selftests (static)       |

* libbpf dynamically linked (libbpf from OS)

[...]

PR #32, opened 2021-07-07
Change BPFMap method `Update` to unsafe.Pointer

grantseltzer:
This makes a breaking change to the libbpfgo API in which
the BPFMap.Update() method takes two unsafe.Pointer instead
of two interface{}. This is so that libbpfgo does not have
to know or care about the type of the passed value. If bpf
and libbpf can support the datatype, so should libbpfgo.

It's important to note that it would then be on the user to
correctly pass arrays/slices (reference must be passed to the
first element, not the array/slice itself).

This also adds a selftest but I haven't added it to any build
script/Makefile, as I'm waiting for #28 to merge, then I will
rebase and update accordingly. 

We should heavily consider the implications of this breaking
API change.

cc @derekparker 

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -472,29 +472,35 @@ func (b *BPFMap) ValueSize() int {
 	return int(C.bpf_map__value_size(b.bpfMap))
 }
 
-func (b *BPFMap) GetValue(key interface{}) ([]byte, error) {
-	keyPtr, err := GetUnsafePointer(key)
-	if err != nil {
-		return nil, fmt.Errorf("failed to lookup value in map %s: unknown key type %T", b.name, key)
-	}
-
+// GetValue takes a pointer to the key which is stored in the map.
+// It returns the associated value as a slice of bytes.
+// All basic types, and structs are supported as keys.
+//
+// NOTE: Slices and arrays are also supported but special care
+// should be taken as to take a reference to the first element
+// in the slice or array instead of the slice/array itself, as to
+// avoid undefined behavior.
+func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
 	value := make([]byte, b.ValueSize())
 	valuePtr := unsafe.Pointer(&value[0])
 
-	errC := C.bpf_map_lookup_elem(b.fd, keyPtr, valuePtr)
+	errC := C.bpf_map_lookup_elem(b.fd, key, valuePtr)
 	if errC != 0 {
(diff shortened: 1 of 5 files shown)


libbpfgo.go
```diff
@@ -501,17 +501,23 @@ func (b *BPFMap) DeleteKey(key interface{}) error {
...
+//
+// For example:
+//
+//  key := 1
+//  value := []byte{'a', 'b', 'c'}
+//  keyPtr := unsafe.Pointer(&key)
+//  valuePtr := unsafe.Pointer(&value[0])
+//  bpfmap.Update(keyPtr, valuePtr)
+//
+func (b *BPFMap) Update(key, value unsafe.Pointer) error {
```
derekparker, line 519: Should the rest of the `BPFMap` also be updated to reflect this change? E.g. `GetValue` and `DeleteKey`? That way you could delete the entire `GetUnsafePointer` function (note this will also be a breaking API change as that method is current exported).

I assume there aren't any bad implications from a consumer trying to use a pointer to a struct or an array/slice as a key, correct (or anything that specifically wasn't allowed before)?

derekparker:
Well, technically this lib is still pre 1.0 so breaking changes can happen! Alternative could be a new method such as `UpdateUnsafe` or something along those lines.

libbpfgo.go
yanivagman: > Should the rest of the BPFMap also be updated to reflect this change? E.g. GetValue and DeleteKey? That way you could delete the entire GetUnsafePointer function 

+1

> I assume there aren't any bad implications from a consumer trying to use a pointer to a struct or an array/slice as a key, correct (or anything that specifically wasn't allowed before)?

With this change, the same semantics as of libbpf are applied. This means that the kernel will try to read from the given pointer the same type that was defined for the map. It is now the responsibility of the consumer to make sure that the pointer given to this function is valid and points to the correct data type, or else, bad things might happen.

grantseltzer: Updated GetValue and DeleteKey accordingly.

grantseltzer:
I think since it's pre-1.0 (for both libbpfgo and libbpf) we should just break the API, but we should make an announcement, cut another release, and  reach out to people we know are using libbpfgo as best as we can to make sure it's well known.

derekparker:
I've tested this change locally with some work I'm doing with Delve and this change now allows me to pass a struct between the userspace program and the eBPF program using an eBPF map and everything seems to work great.

yanivagman (approved):

Issue #35, opened 2021-07-09
Fix broken selftests


grantseltzer:
In #32 I made breaking API changes and failed to update the selftests accordingly. After #28 merges we should follow up immediately with fixing the selftests. Additionally we need to make sure the new map_update selftest is being run as part of the same selftest run target.

rafaeldtinoco:
#28 has been merged and map-update gives us:

```
$ make -C selftest/map-update run-static
make: Entering directory '/home/rafaeldtinoco/work/sources/ebpf/libbpfgo/selftest/map-update'
make -C /home/rafaeldtinoco/work/sources/ebpf/libbpfgo libbpfgo-static
make -C /home/rafaeldtinoco/work/sources/ebpf/libbpfgo vmlinuxh
make[1]: Entering directory '/home/rafaeldtinoco/work/sources/ebpf/libbpfgo'
make[1]: Entering directory '/home/rafaeldtinoco/work/sources/ebpf/libbpfgo'
mkdir -p ./output
mkdir -p ./output
mkdir -p ./output/libbpf
INFO: generating ./output/vmlinux.h from /sys/kernel/btf/vmlinux
CC="gcc" CFLAGS="-g -O2 -Wall -fpie" LD_FLAGS="" \
   make -C /home/rafaeldtinoco/work/sources/ebpf/libbpfgo/libbpf/src \
	BUILD_STATIC_ONLY=1 \
	OBJDIR=/home/rafaeldtinoco/work/sources/ebpf/libbpfgo/output/libbpf \
	DESTDIR=/home/rafaeldtinoco/work/sources/ebpf/libbpfgo/output \
	INCLUDEDIR= LIBDIR= UAPIDIR= install
make[2]: Entering directory '/home/rafaeldtinoco/work/sources/ebpf/libbpfgo/libbpf/src'
  MKDIR    staticobjs
  INSTALL  bpf.h libbpf.h btf.h libbpf_common.h libbpf_legacy.h xsk.h bpf_helpers.h bpf_helper_defs.h bpf_tracing.h bpf_endian.h bpf_core_read.h skel_internal.h
  CC       bpf.o
  CC       btf.o
  CC       libbpf.o
  CC       libbpf_errno.o
  CC       netlink.o
  CC       nlattr.o
  CC       str_error.o
  CC       libbpf_probes.o
  CC       bpf_prog_linfo.o
  CC       btf_dump.o
  CC       xsk.o
  CC       ringbuf.o
[...]

PR #36, opened 2021-07-12
Fix iterators selftest to use updated unsafe Pointer API

grantseltzer:
Handles #35 

### DIFF
--- a/selftest/iterators/main.go
+++ b/selftest/iterators/main.go
@@ -17,6 +17,8 @@ const (
 	checked     uint32 = 2
 )
 
+var one = 1
+
 func main() {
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
It works!

Before:
```
# command-line-arguments
./main.go:40:23: cannot use i (type uint32) as type unsafe.Pointer in argument to numbers.Update
./main.go:40:23: cannot use added (type uint32) as type unsafe.Pointer in argument to numbers.Update
make: *** [Makefile:62: main-static] Error 2
```

After: good

rafaeldtinoco (approved):

PR #41, opened 2021-07-26
New helper (getter) functions and some refactoring

geyslan:
This originated from Discussion https://github.com/aquasecurity/libbpfgo/discussions/40

It adds the following functions.

For BPFMap:

```go
GetFd()
GetName()
GetPinPath()
IsPinned()
GetModule()
```

For BPFProg:

```go
GetName()
GetModule()
```

It also does some refactoring to make reuse of the already got libbpf map pointer: `bpfMap *C.struct_bpf_map`.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -378,11 +378,8 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 }
 
 func (b *BPFMap) Pin(pinPath string) error {
-	cs := C.CString(b.name)
 	path := C.CString(pinPath)
-	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
-	errC := C.bpf_map__pin(bpfMap, path)
-	C.free(unsafe.Pointer(cs))
+	errC := C.bpf_map__pin(b.bpfMap, path)
 	if errC != 0 {
 		return fmt.Errorf("failed to pin map %s to path %s", b.name, pinPath)
 	}
@@ -390,11 +387,8 @@ func (b *BPFMap) Pin(pinPath string) error {
 }
 
 func (b *BPFMap) Unpin(pinPath string) error {
-	cs := C.CString(b.name)
 	path := C.CString(pinPath)
-	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
-	errC := C.bpf_map__unpin(bpfMap, path)
-	C.free(unsafe.Pointer(cs))
+	errC := C.bpf_map__unpin(b.bpfMap, path)
 	if errC != 0 {
 		return fmt.Errorf("failed to unpin map %s from path %s", b.name, pinPath)
 	}
@@ -402,11 +396,8 @@ func (b *BPFMap) Unpin(pinPath string) error {
 }
 
 func (b *BPFMap) SetPinPath(pinPath string) error {
-	cs := C.CString(b.name)
 	path := C.CString(pinPath)
-	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
-	errC := C.bpf_map__set_pin_path(bpfMap, path)
-	C.free(unsafe.Pointer(cs))
+	errC := C.bpf_map__set_pin_path(b.bpfMap, path)
 	if errC != 0 {
 		return fmt.Errorf("failed to set pin for map %s to path %s", b.name, pinPath)
 	}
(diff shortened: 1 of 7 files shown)


geyslan:
I split the functions commits leaving the `GetModule()` in the https://github.com/aquasecurity/libbpfgo/pull/41/commits/468d2b2e0185e60b6087c23d4a2a85254f00bf7e, since this getter might not be welcome.

libbpfgo.go
```diff
@@ -440,6 +440,33 @@ func (b *BPFMap) GetMaxEntries() uint32 {
...
+}
+
+func (b *BPFMap) GetPinPath() string {
+	cs := C.CString(b.name)
+	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
+	pinPathGo := C.GoString(C.bpf_map__get_pin_path(bpfMap))
+	C.free(unsafe.Pointer(cs))
+	return pinPathGo
+}
+
```
rafaeldtinoco, line 462: Could you add a selftest for GetPinPath() and IsPinned() ?
Something very simple, making sure they work and don't regress ?

> Note: You can clone a simple examples from selftest/ directory.

geyslan: @rafaeldtinoco, done. Thanks a million.

rafaeldtinoco: I like it. I'll let sit here for awhile (to see if others have any comment). Will run the selftest together with others and if all good will commit. Thank U for the contribution.

libbpfgo.go
```diff
@@ -440,6 +440,37 @@ func (b *BPFMap) GetMaxEntries() uint32 {
...
+	return b.name
+}
+
+func (b *BPFMap) GetModule() *Module {
+	return b.module
+}
+
+func (b *BPFMap) GetPinPath() string {
+	cs := C.CString(b.name)
+	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
```
yanivagman, line 457: This step is unnecessary as you can get it from the bpfMap field of a BPFMap

libbpfgo.go
```diff
@@ -440,6 +440,37 @@ func (b *BPFMap) GetMaxEntries() uint32 {
...
+	pinPathGo := C.GoString(C.bpf_map__get_pin_path(bpfMap))
+	C.free(unsafe.Pointer(cs))
+	return pinPathGo
+}
+
+func (b *BPFMap) IsPinned() bool {
+	cs := C.CString(b.name)
+	bpfMap := C.bpf_object__find_map_by_name(b.module.obj, cs)
```
yanivagman, line 465: same

yanivagman (commented):
Overall LGTM, just a small comment to save some cycles

libbpfgo.go
geyslan: Yes we can get it inside the BPFMap, indeed. I've noticed that other functions do the same, getting again the same reference:

https://github.com/aquasecurity/libbpfgo/blob/b1eb1b9a6b20b60d0cc0ebfb374ec08684d1682e/libbpfgo.go#L380-L383

https://github.com/aquasecurity/libbpfgo/blob/b1eb1b9a6b20b60d0cc0ebfb374ec08684d1682e/libbpfgo.go#L392-L395

And so on.

Is there a specific reason for those?

yanivagman: We should change these as well.

> Would it be related to maps that could be changed by other codepaths

Changed how? bpfMap (bpf_map) is used by libbpf as a reference to a map so that it can use the bpf syscall with. It is mostly the fd of the struct that is used for these operations. I don't see a problem with working with that for the lifecycle of the BPFMap object.

geyslan: > We should change these as well.

Grand, I can do that in this PR or a separated one. Please let me know what you do prefer.

> Changed how?

Other software using direct access (bpf syscall) could pin or unpin or set a different pinpath to one running map. Would be the libbpf maps fields automatically updated in this case?

yanivagman: > Grand, I can do that in this PR or a separated one. Please let me know what you do prefer.

You can add it to this PR if you like, not a must though.

> Other software using direct access (bpf syscall) could pin or unpin or set a different pinpath to one running map. Would be the libbpf maps fields automatically updated in this case?

Even when other programs pin/unpin/change pin path, as long as you have a reference to a map, you can use it to do other operations on it. Here is the code in libbpf that is being called from here: https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L8509 - as you can see, the bpfMap reference is enough to get the relevant data. So in our context, I don't think it is a problem (although you are right that it may be a problem for the validity of some other fields, like if the map is pinned or not, but we don't care about it here).

geyslan: > You can add it to this PR if you like, not a must though.

Done. Anything, please let me know. :+1:

rafaeldtinoco (approved):
```
$ make libbpfgo-static-test
ok  	github.com/aquasecurity/libbpfgo	0.530s

$ make selftest-static
ok

$ ls -1 selftest/map-pin*/main-static
selftest/map-pin-info/main-static

$ pwd
.../ebpf/libbpfgo/selftest/map-pin-info

$ make run
[*] SUCCESS: all good
```

libbpfgo.go
rafaeldtinoco: Nice catch on bpfMap. All tests seem fine, I approved it. @yanivagman let me know if you're good so I can merge, or feel free to merge it yourself once u think is good. Thanks!

rafaeldtinoco: I'll go ahead and merge this if you don't mind, we can always change it if needed =).

PR #44, opened 2021-07-27
Makefile: download libbpf submodule when not found

geyslan:
Making contribution easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ ARCH := $(subst x86_64,amd64,$(ARCH))
 
 BTFFILE = /sys/kernel/btf/vmlinux
 BPFTOOL = $(shell which bpftool || /bin/false)
+GIT = $(shell which git || /bin/false)
 VMLINUXH = $(OUTPUT)/vmlinux.h
 
 # libbpf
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
This way an offline build would fail, wouldn't it ? Sometimes we may need to rsync the git tree to an offline node to test building, for example. Maybe to check if directory is empty and then try to update it ?

Not sure about the submodule managing by Makefile also, @grantseltzer what is your take on this ?

geyslan:
>  Maybe to check if directory is empty and then try to update it ?

The rule already does something like that since `LIBBPF_SRC = $(abspath ./libbpf/src)`. So the rule only runs if `./libbpf/src` doesn't exist.

```Makefile
$(LIBBPF_SRC):
ifeq ($(wildcard $@), )
	echo "INFO: updating submodule 'libbpf'"
	$(GIT) submodule update --init --recursive
endif
```

rafaeldtinoco:
> ifeq ($(wildcard $@), )

Sorry, missed that. Okay then, I'm ok with this.

grantseltzer (approved):
LGTM! Thanks so much @geyslan !

PR #38, opened 2021-07-12
examples: Add tcpconnect as a libbpfgo example

rafaeldtinoco:
$ make -C examples/tcpconnect

will build:

- tcpconnect-static
- tcpconnect-dynamic
- tcpconnect-go-static
- tcpconnect-go-dynamic

as an example on how to use libbpfgo (or libbpf) to probe kernel
function "tcp_connection" and get data from it.

PR #45, opened 2021-07-27
examples: Add tcpconnect as a libbpfgo example

rafaeldtinoco:
$ make -C selftest/tcpconnect
$ make -C selftest/tcpconnect run

will build:

- tcpconnect-static
- tcpconnect-dynamic
- tcpconnect-go-static
- tcpconnect-go-dynamic

as an example on how to use libbpfgo (or libbpf) to probe kernel
function "tcp_connection" and get data from it. It will also use
this example as a selftest for a simple TCP connect().

### DIFF
--- a/selftest/tcpconnect/README.md
+++ b/selftest/tcpconnect/README.md
@@ -1,0 +1,281 @@
+## libbpfgo: tcpconnect example
+
+### Introduction
+
+The idea behind this test directory is to document how you can migrate from a C eBPF based code, using [libbpf](https://github.com/libbpf/libbpf), to a pure golang eBPF code, using [libbpfgo](https://github.com/aquasecurity/libbpfgo). You may also read the Makefile to better understand how to build static and dynamic Go binaries with [libbpfgo](https://github.com/aquasecurity/libbpfgo).
+
+The directory structure is this:
+
+* main.bpf.o: the eBPF program loaded to the kernel
+* main.go: the Go version of the userland program
+* main.c: the C version of the userland progam
+
+Both main.go and main.c programs should behave similarly, if not identically.
+
+There are better places to learn eBPF, like:
+
+1. https://ebpf.io/what-is-ebpf
+1. https://ebpf.io/what-is-ebpf#introduction-to-ebpf
+1. https://ebpf.io/what-is-ebpf#maps
+1. https://ebpf.io/what-is-ebpf#why-ebpf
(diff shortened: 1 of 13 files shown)


rafaeldtinoco:
This is the old https://github.com/aquasecurity/libbpfgo/pull/38 which I accidentally lost after a force push in my local repository. I'm bringing it back now. 

@grantseltzer thanks for reviewing this, it was missing a Readme.MD file only.

selftest/tcpconnect/tcpclient
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sleep 5 ; echo "connected" | nc -q 2 127.0.0.1 12345 > /dev/null
```
grantseltzer, line 3: On fedora the `nc` that's packaged does not have a `-q` flag. The only similar options I see:

`--send-only            Only send data, ignoring received; quit on EOF`
`w, --wait <time>          Connect timeout`

selftest/tcpconnect/tcpserver
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+nc -F -l -p 12345 > /dev/null
```
grantseltzer, line 3: Same with `-F`. Looks like there's an openbsd, gnu, and 'classic' netcats. A licensing issue.

selftest/tcpconnect/tcpclient
rafaeldtinoco: Hum, I'll change that usage then... did not expect that.

selftest/tcpconnect/tcpserver
rafaeldtinoco: Same here.

rafaeldtinoco:
@grantseltzer thanks. I believe those are fixed for Fedora now.

grantseltzer (approved):
LGTM :-)

selftest/tcpconnect/README.md
```diff
@@ -0,0 +1,281 @@
...
+        struct in6_addr daddr6; // destination address (IPv6)
+        u8   thesource;         // I am the one originating packet
+};
+```
+
+This means that both, the eBPF program, and the userland code, need to know about this structure, so they can communicate.
+
+> Note: there are different types of events which an eBPF program can be attached to (links) and different types of maps that can be used to exchange data. I'm not entering that topic here and this example uses simple ones.
+
+From the snippet bellow:
```
grantseltzer, line 63: here too

grantseltzer (approved):

PR #46, opened 2021-07-27
fix potential memory leaks

yanivagman:


rafaeldtinoco (approved):
Yes, I saw that and questioned myself. Did not ask anything because thought you saw something else. Lesson learned and +1.

Issue #39, opened 2021-07-22
Versioning libbpfgo


yanivagman:
As libbpfgo depends on libbpf and there are features in libbpfgo that require some minimal version of libbpf, we should add the version of libbpf to be part of the version of libbpfgo.
This can be, for example:
v0.2-libbpf_0.4

PR #47, opened 2021-08-02
Add section in README about release versioning

grantseltzer:
This describes @yanivagman's idea in #39.  I would immediately follow up this PR merge with the cutting of a new release based on the scheme described here.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -101,6 +101,9 @@ libbpfgo does not yet have a regular schedule for cutting releases. There has no
 - __Patch releases__ are cut to incorporate important individual or groupings of bug fixes.
 - __libbpf support numbering__ indicates the _minimum_ required libbpf version that must be linked in order to ensure libbpfgo compatibility. For example, `v0.2.1-libbpf_0.4.0` means that version 0.2.1 of libbpfgo requires v0.4.0 or newer of libbpf.
 
+*Note*: some distributions might have local changes to their libbpf package and their version might include backports and/or fixes differently than upstream versions. In those cases we recommend that libbpfgo is used statically compiled.
+
+
 ## Learn more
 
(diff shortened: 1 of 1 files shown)


rafaeldtinoco:
> Major releases are cut when backwards compatibility is broken or major milestones are completed, such as reaching parity with libbpf's API.

Agreed.

> Minor releases are cut to incorporate new support for libbpf APIs.

Yes. If we follow a 1:1 feature with libbpf like we're doing, then this will work.

> Patch releases are cut to incorporate important individual or groupings of bug fixes.

Yes.

> libbpf support numbering indicates the minimum required libbpf version that must be linked in order to ensure libbpfgo compatibility. For example, v0.2.1-libbpf_0.4.0 means that version 0.2.1 of libbpfgo requires v0.4.0 or newer of libbpf.

Perfect.

--

Can we add a disclaimer like (and feel free to rephrase as English is better on your side):

Do note that some distributions might have local changes to their libbpf package and their version might include backports and/or fixes differently than upstream versions.

grantseltzer:
> Can we add a disclaimer like (and feel free to rephrase as English is better on your side):
> 
> Do note that some distributions might have local changes to their libbpf package and their version might include backports and/or fixes differently than upstream versions. In those cases we recommend that libbpfgo is used statically compiled.

Great idea, and your english was perfect. Added.

> And then I can do a simple write on how to clone libbpfgo and write a oneliner and have it compiled with dynamic libbpfgo dependency and with static libbpfgo dependency.
> 
> What do you think ?

Sounds great!

rafaeldtinoco (approved):
LGTM

grantseltzer:
I pushed another commit when I realized I had some redundant information in my previous one. Also added a TOC. 

Also want to note that I confirmed having '-' or '_' doesn't break anything with versioning go modules.

PR #49, opened 2021-08-05
Fix init_perf_buf error handling

josedonizetti:
While investigating [different capabilities](https://github.com/aquasecurity/tracee/issues/747) to avoid `sudo`,  I get

```
./tracee-ebpf 
libbpf: failed to mmap perf buffer on cpu #7: Operation not permitted
TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
[signal SIGSEGV: segmentation violation code=0x1 addr=0x3f pc=0x713b73]

runtime stack:
runtime.throw(0x7e9fe6, 0x2a)
	/usr/local/go/src/runtime/panic.go:1117 +0x72
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:718 +0x2e5
...
```

The PR fixes error handling to properly stop the service instead of starting it without the perf buffers.

```
./tracee-ebpf 
libbpf: failed to mmap perf buffer on cpu #7: Operation not permitted
Failed to initialize perf buffer!
2021/08/05 14:27:37 error creating Tracee: error initializing file_writes perf map: failed to initialize perf buffer
```

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -70,7 +70,7 @@ struct perf_buffer * init_perf_buf(int map_fd, int page_cnt, uintptr_t ctx) {
     pb_opts.lost_cb = perfLostCallback;
     pb_opts.ctx = (void*)ctx;
     pb = perf_buffer__new(map_fd, page_cnt, &pb_opts);
-    if (pb < 0) {
+    if (libbpf_get_error(pb)) {
         fprintf(stderr, "Failed to initialize perf buffer!\n");
         return NULL;
     }


rafaeldtinoco:
Hello @josedonizetti , thanks for the commit. While the change is appropriate...

I'm curious how you got into that situation. If it was because of a code change AND a capability drop test (using capsh for example), done simultaneously, would you mind sharing the reproducer ?

Currently if I drop cap_sys_admin I get insufficient privileges to run. If I drop cap_bpf, cap_sys_admin allows me to run bpf (likely for backwards compatibility). 

My thought: if we are breaking something because of a development effort maybe we should do all changes together.

Again, obviously the fix is appropriate, just wanted to get documented how to get there.

rafaeldtinoco:
BTW, I see that your suggestion is against libbpfgo, so I guess it is right to be done separately anyway! Thank you!

josedonizetti:
@rafaeldtinoco  To simulate it is a mix of changes, I started by changing [`tracee-ebpf`](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/main.go#L949), my change was in order to address the [issue](https://github.com/aquasecurity/tracee/issues/747).

If the kernel was at least 5.8, I was testing for `CAP_BPF`, `CAP_PERFMON`, and `CAP_NET_ADMIN` instead of testing `CAP_SYS_ADMIN`. 

With this change in place, I build a new `tracee-ebpf` and changed it capabilities: `sudo setcap cap_net_admin,cap_ebpf,cap_perfmon=pe ./dist/tracee-ebpf`

Then trying to execute the command without sudo triggered the error reported.

rafaeldtinoco:
Thank you for the feedback!

rafaeldtinoco (approved):
Correct way of dealing with ERR_PTR() from __perf_buffer_new().

Issue #51, opened 2021-08-14
Program GetFd() return type


geyslan:
Should we return a Go int instead of C.int?

https://github.com/aquasecurity/libbpfgo/blob/871dc676329da8991fadc7395f34ad81daef8898/libbpfgo.go#L650-L652

Like this:

https://github.com/aquasecurity/libbpfgo/blob/871dc676329da8991fadc7395f34ad81daef8898/libbpfgo.go#L466-L468

rafaeldtinoco:
I think so, yes (at least for conformity). And tracee will have to fix some usages:

![image](https://user-images.githubusercontent.com/7395852/129466714-5570bb95-0608-4a4a-b0fa-dc0930006a9d.png)

geyslan:
If you agree I can raise a PR for both.

yanivagman:
Agree. Let's change this to return an int like BPFMap.
@rafaeldtinoco I'm not sure we need to change something in Tracee.
Anyways, let's make this change first and then see if there are any required changes.

rafaeldtinoco:
Sure, let's consider libbpfgo and tracee independent.

Currently I'm checking libbpfgo changes for tracee (specially GetMap, UpdateMap) and I can see that the new implementations of:

- bpfMap.GetValue()
- bpfMap.DeleteKey()
- bpfMap.Update()

are the **only** functions having unsafe.Pointer() as arguments: Users have to specify C types as arguments instead of relying in old interface{} arguments (that had issues). All the rest of functions seem to have go types as arguments. 

> bpfMap.GetValue() correctly returns go type: ([]byte, error).

We may need to review return types, so all functions return Go types. Example:

```go
func (p *BPFProg) GetType() uint32 {
	return C.bpf_program__get_type(p.prog)
}
```

should maybe `return C.uint(C.bpf_program__get_type(p.prog)` instead, and so on.

geyslan:
I've detected `BPFProg.GetFd()` as the only one returning a non go type.

PR #52, opened 2021-08-16
libbpfgo: change BPFProg.GetFd() return to int Go type

geyslan:
For conformity, make available for consumer only Go types.

Fixes: #51


rafaeldtinoco (approved):
LGTM.

PR #53, opened 2021-08-17
helpers: add initial btfinfo

rafaeldtinoco:
During https://github.com/aquasecurity/tracee/pull/895 discussions we agreed that the merge being done there would be done into libbpfgo and added as helper.

### DIFF
--- a/helpers/btfinfo.go
+++ b/helpers/btfinfo.go
@@ -1,0 +1,116 @@
+package helpers
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"golang.org/x/sys/unix"
+	"os"
+	"strings"
+)
+
+func fileExists(path string) bool {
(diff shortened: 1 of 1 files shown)


helpers/btfinfo.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return false
+}
+
+// fileGrepUnique: returns first 'needle' string found in a file 'path'.
+func fileGrepUnique(path string, needle string) string {
+	file, _ := os.Open(path)
+	defer func(file *os.File) {
+		_ = file.Close()
+	}(file)
```
yanivagman, line 24: Why not simply deferring file.Close()?

rafaeldtinoco: this was added automatically by goland ... argh

yanivagman (approved):
LGTM

PR #59, opened 2021-08-28
checking if new operation in newbtfino succeeded

mtcherni95:
We now check if new operation succeeded. so that in case it didn't we don't panic by calling DiscoverDistro().

The user of the NewBTFInfo function is responsible to check if the returned error is not nil.


rafaeldtinoco (approved):
Thanks for the fix and for quick discussion about return &{} and new().

PR #62, opened 2021-08-30
add parser for mknod mode argument

roikol:


yanivagman (commented):
How is this function any different from ParseInodeMode?

rafaeldtinoco (approved):
LGTM

yanivagman (changes requested):
ParseInodeMode() should be good enough, I think.
Anyways, why not using uint32 for mode_t?

roikol:
after talking with @yanivagman - ParseInodeMode() is good for my needs, so this new function ParseMknodMode() is not needed. closing the PR.

PR #63, opened 2021-08-31
kernel_features: fix kconfig file read logic for procfs config.gz

rafaeldtinoco:


rafaeldtinoco:
@grantseltzer my previous commits to helpers/kernel_features.go have broken /proc/config.gz support (I'm testing tracee in all environments now - https://github.com/aquasecurity/tracee/issues/971 - and that is how I got this regression). It is not affecting tracee until we change its go.mod so we're good, after this fix, to bump libbpfgo version in tracee.

> We will need to raise libbpfgo version in tracee because of this change and I have a commit ready:

443f6a5 - (HEAD -> rafaeldtinoco) tracee: move MissingKernelConfigOptions to libbpfgo helper (2 hours ago) <Rafael David Tinoco>

Could you review this for me ? Thank you

helpers/kernel_features.go
```diff
@@ -180,6 +181,18 @@ type KernelConfig struct {
 
 // InitKernelConfig inits external KernelConfig object
 func InitKernelConfig() (*KernelConfig, error) {
+	config := KernelConfig{}
+
+	// fastpath: check config.gz in procfs first
+	if _, err1 := os.Stat("/proc/config.gz"); err1 == nil {
```
yanivagman, line 187: Is this required here? You already do this stat in initKernelConfig

yanivagman (approved):

PR #64, opened 2021-09-01
add parser for kernel_read_file_id enum

roikol:
kernel_read_file_id in an enum that tells what is being read into the kernel. if merged - it'll be used by security_kernel_read_file event in tracee-ebpf.
this enum was changed between kernel versions.

### DIFF
--- a/helpers/argumentParsers.go
+++ b/helpers/argumentParsers.go
@@ -635,9 +635,7 @@ func UnameRelease() string {
 		buf[i] = byte(b)
 	}
 	ver := string(buf[:])
-	if i := strings.Index(ver, "\x00"); i != -1 {
-		ver = ver[:i]
-	}
+	ver = strings.Trim(ver, "\x00")
 	return ver
 }
 
--- a/helpers/argumentParsers.go
+++ b/helpers/argumentParsers.go
@@ -2,9 +2,11 @@ package helpers
 
 import (
 	"encoding/binary"
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
+	"syscall"
 )
 
 // ParseInodeMode parses the `mode` bitmask argument of the `mknod` syscall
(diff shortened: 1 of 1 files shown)


helpers/argumentParsers.go
```diff
@@ -621,3 +623,95 @@ func ParseBPFCmd(cmd int32) string {
...
+	var uname syscall.Utsname
+	if err := syscall.Uname(&uname); err != nil {
+		return ""
+	}
+	var buf [65]byte
+	for i, b := range uname.Release {
+		buf[i] = byte(b)
+	}
+	ver := string(buf[:])
+	if i := strings.Index(ver, "\x00"); i != -1 {
```
rafaeldtinoco, line 638: strings.Trim() would do that in 1 line.

helpers/argumentParsers.go
```diff
@@ -621,3 +623,95 @@ func ParseBPFCmd(cmd int32) string {
...
+		ver = ver[:i]
+	}
+	return ver
+}
+
+func ParseKernelReadFileId(id int32) (string, error) {
```
rafaeldtinoco, line 644: I feel this needs some explanation on why it should exist, despite being just a helper.

helpers/argumentParsers.go
```diff
@@ -621,3 +623,95 @@ func ParseBPFCmd(cmd int32) string {
 
 	return res
 }
+
+// UnameRelease gets the version string of the current running kernel
+func UnameRelease() string {
```
rafaeldtinoco, line 628: You're bringing this from tracee ? I'm also using similar in BTFinfo on its new version:

https://github.com/rafaeldtinoco/libbpfgo/commit/ae3c034dadf449ef36b7a6701520522db50473fc

 Maybe we should put this in a common file to all helpers.

rafaeldtinoco (changes requested):

helpers/argumentParsers.go
roikol: yeah, i took it from tracee. 
i agree - this should be a common helper in libbfgo, and tracee should consume that wherever needed.

helpers/argumentParsers.go
roikol: no problem, i'll a few lines to explain.

roikol: i've added a small explanation - is that good enough or should i elaborate more?

helpers/argumentParsers.go
yanivagman: Agree. We should move both UnameRelease() and kernelIsAtLeast() to helpers

helpers/argumentParsers.go
yanivagman: This is not the place for such a function.
The helpers package is intended for a user of libbpfgo, which might need to parse known types used by different Linux system calls. This enum, however, is an internal kernel type, which is too specific to be included here

yanivagman (changes requested):

helpers/argumentParsers.go
roikol: this enum is used in the LSM hook `security_kernel_read_file`  which the user may use. that's why i thought this is a good place for this helper. see: https://elixir.bootlin.com/linux/v5.14/source/fs/kernel_read_file.c#L20

helpers/argumentParsers.go
roikol: should it be as part of this PR?

rafaeldtinoco: I vote for a simple PR doing the needed moves first and then this one rebased on top of that. That because I'm also using Uname function in my new OSInfo class (renaming BTFInfo class). I can also rebase on top of this new PR.

helpers/argumentParsers.go
rafaeldtinoco: I had the same concerns as @yanivagman thus my question to 'explain' it further.

roikol: ok, so i'll remove this helper out of libbpgo.

roikol:
ok, since we are moving out the parser func, and decided to create a PR for the kernel version helpers - i'm closing this PR, and will open a new one for the kernel version helpers.
thank you

Issue #61, opened 2021-08-29
BTF: support providing external kconfig


yanivagman:
When running libbpf in an environment where BTF is enabled, it requires having kconfig data which is read from either `/proc/config.gz `or `/boot/config-xxx`.
When working from a container in an environment where kconfig is located under /boot and it is not mounted, libbpf might fail to load the BPF object. To avoid this situation, we can expand libbpfgo to support providing extra kconfig values and give it as an argument to libbpf. See https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L96

rafaeldtinoco:
I recently changed KernelConfig in libbpfgo. There is a regression for /proc/config.gz (happening in archlinux) for that version. I'll fix that and work in this issue together:

https://github.com/aquasecurity/libbpfgo/pull/57
and
https://github.com/aquasecurity/libbpfgo/pull/58

rafaeldtinoco:
@yanivagman I'm almost done with the changes that allows us to specify KCONFIG and OSRELEASE files through environment. I think they're nice BUT wanted to inform you that libbpf still complains about the missing kconfig file because it loads both: the given kconfig and the system kconfig. Given kconfig is not meant to replace the system one (we might need to propose a change to them):

```C
	if (need_config && extra_kconfig) {
	    fprintf(stdout, "TEMPORARY 02: %s\n", extra_kconfig);
		err = bpf_object__read_kconfig_mem(obj, extra_kconfig, kcfg_data);
		if (err)
			return -EINVAL;
		need_config = false;
		for (i = 0; i < obj->nr_extern; i++) {
			ext = &obj->externs[i];
			if (ext->type == EXT_KCFG && !ext->is_set) {
				need_config = true;
				break;
			}
		}
	}
	if (need_config) {
		err = bpf_object__read_kconfig_file(obj, kcfg_data);
		if (err)
			return -EINVAL;
	}
```

my output:

```
$ sudo TRACEE_KCONFIG_FILE=/boot/config-5.14.0-rc5+.orig TRACEE_OSRELEASE_FILE=/etc/os-release.orig ./dist/tracee-ebpf --debug --trace uid=1000 --trace pid=new --trace event=execve --trace event=cap_capable --output option:detect-syscall
OSInfo: VERSION: "21.10 (Impish Indri)"
OSInfo: VERSION_CODENAME: impish
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
[...]

Issue #60, opened 2021-08-29
helpers: Discover BTFinfo from inside a container


yanivagman:
We currently look for the distro info in `/etc/os-release` (https://github.com/aquasecurity/libbpfgo/blob/main/helpers/btfinfo.go#L91).
When we run from inside a container, this file may represent a different distro, which will give us the wrong BTFinfo.
We need to be able to support an external /etc/os-release file as an argument, so it will be possible to point to a mounted file instead of the hardcoded /etc/os-release

rafaeldtinoco:
TODO: add an option to override whatever the discovery process is doing.

rafaeldtinoco:
My recent pull request allows us to bind mount /etc/os-release to something like /etc/os-release-host in the container and feed that to tracee using TRACEE_OSRELEASE_FILE environment variable. If not bound, we can also opt to create a fake os-release file with the host information and provide that as well. 

> I'm going to add cmdline support for providing ID and VERSION_ID values from os-release directly. If those are provided then tracee won't even read /etc/os-release OR TRACEE_OSRELEASE_FILE.

PR #67, opened 2021-09-01
osinfo and kernel_features (kernel_config) refactoring and improvements

rafaeldtinoco:
osinfo and kernel_features (kernel_config) refactoring and improvements

- helpers/kernel_features: allow kconfig override 
- helpers/btfinfo: rename to osinfo and improve API 
- helpers/common: functions common to all helpers
- helpers/kernel_config: kernel_features refactoring 
- helpers/osinfo: osinfo refactoring for env variable 

Reference: https://github.com/aquasecurity/libbpfgo/issues/60
Reference: https://github.com/aquasecurity/libbpfgo/issues/61


rafaeldtinoco:
Note: I have not done the cmdline OSInfo options to override if no /etc/os-release is found. Will do in a subsequent PR. I wanted this to be reviewed and merged first (and then to sync tracee with this change so its stable and users can use the kernel version comparison function).

helpers/osinfo.go
```diff
@@ -0,0 +1,270 @@
...
+// This is a less complete approach than NewOSInfo call, recommended
+func NewOSInfoRelease() (*OSInfo, error) {
+	info := OSInfo{}
+
+	if err := info.DiscoverRelease(); err != nil {
+		return nil, err
+	}
+
+	return &info, nil
+}
```
yanivagman, line 189: I'm not sure these two are needed.
It is a little confusting that `NewOSInfoRelease()` doesn't populate all the fields in the struct, while `NewOSInfo()` does. Anyways, OSInfo struct is exported, and all what these functions do is to either call `DiscoverDistro()` or `DiscoverRelease()`, so maybe it's better to remove them?

yanivagman (approved):
One comment to remove "factory" functions.
Otherwise LGTM

helpers/osinfo.go
```diff
@@ -0,0 +1,270 @@
...
+)
+
+type OSReleaseField uint32
+
+func (o OSReleaseField) String() string {
+	return OSReleaseFieldToString[o]
+}
+
+// StringToOSReleaseField is a map of os-release file fields
+var StringToOSReleaseField = map[string]OSReleaseField{
```
itaysk, line 73: please reconsider what needs to be exported and what not (across the PR)

helpers/osinfo.go
```diff
@@ -0,0 +1,270 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+)
+
+type OSReleaseID uint32
```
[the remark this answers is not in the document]
rafaeldtinoco: We've discussed this already in a meeting (the type is being stringified)

helpers/osinfo.go
rafaeldtinoco: Those are all the fields os-release implementation supports. Not something I picked to support, if that is what you meant here. The API supports what https://www.freedesktop.org/software/systemd/man/os-release.html describes as fields in that file.

helpers/osinfo.go
```diff
@@ -0,0 +1,270 @@
...
+
+	ver := string(buf[:])
+	ver = strings.Trim(ver, "\x00")
+
+	return ver, nil
+}
+
+// OSBTFEnabled checks if kernel has embedded BTF vmlinux file
+func OSBTFEnabled() bool {
+	_, err := os.Stat("/sys/kernel/btf/vmlinux")
```
grantseltzer, line 162: Is this the only way of doing this? Does checking the kconfig make sense at all?

helpers/osinfo.go
```diff
@@ -0,0 +1,270 @@
...
+
+	if err := btfi.DiscoverRelease(); err != nil {
+		return err
+	}
+
+	file, _ := os.Open(releaseFilePath)
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+
+	for i := 1; scanner.Scan(); i++ {
```
grantseltzer, line 251:
```suggestion
	for scanner.Scan() {
```

helpers/osinfo.go
rafaeldtinoco: Thats a good question actually @grantseltzer ... There might be other ways. I think that if you read the ELF file and it contains the .BTF section its half way done. You would still miss whether BTF is exported through sysfs, which you would have to check kconfig. Maybe reading kconfig and assuming BTF is enabled is better than trying to see if that file exists (just like any other kconfig feature. If not done by this PR, that is something we should do indeed.

helpers/osinfo.go
rafaeldtinoco: where did that come from ?!?!

grantseltzer: for == while in Go :-)

helpers/osinfo.go
rafaeldtinoco: I'm adding a TODO now .. so OSBTFEnabled() gets a KernelConfig type as parameter and checks if BTF kernel config option is set for the running environment. Dont think we should in this PR as this is being coordinated with tracee and has enough changes already.

helpers/osinfo.go
grantseltzer: I would agree that if we have the String() method, we don't need to export the map.

rafaeldtinoco:
Example of usage within tracee:

```
$ sudo LIBBPFGO_KCONFIG_FILE=/boot/config-5.14.0-rc5+.orig LIBBPFGO_OSRELEASE_FILE=/etc/os-release.orig ./dist/tracee-ebpf --debug --trace uid=1000 --trace pid=new --trace event=execve --trace event=cap_capable --output option:detect-syscall
OSInfo: VERSION_ID: "21.10"
OSInfo: VERSION: "21.10 (Impish Indri)"
OSInfo: VERSION_CODENAME: impish
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
OSInfo: KERNEL_RELEASE: 5.14.0-rc5+
OSInfo: PRETTY_NAME: "Ubuntu Impish Indri (development branch)"
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
```

helpers/osinfo.go
rafaeldtinoco: Oh I see now.

helpers/osinfo.go
```diff
@@ -0,0 +1,231 @@
...
+	OS_VERSION
+	OS_VERSION_ID
+	OS_VERSION_CODENAME
+	OS_BUILD_ID
+	OS_IMAGE_ID
+	OS_IMAGE_VERSION
+	OS_KERNEL_RELEASE // not part of default os-release, but we can use it here to facilitate things
+)
+
+type osReleaseField uint32
```
grantseltzer, line 65: This still needs to be exported for people to be able to use the `.String()` method

helpers/osinfo.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (o osReleaseField) String() string {
+	return osReleaseFieldToString[o]
+}
+
+// StringToOSReleaseField is a map of os-release file fields
```
grantseltzer, line 71: nit: change the case in godoc comment

helpers/osinfo.go
```diff
@@ -0,0 +1,231 @@
...
+	"VERSION":          OS_VERSION,
+	"VERSION_ID":       OS_VERSION_ID,
+	"VERSION_CODENAME": OS_VERSION_CODENAME,
+	"BUILD_ID":         OS_BUILD_ID,
+	"IMAGE_ID":         OS_IMAGE_ID,
+	"IMAGE_VERSION":    OS_IMAGE_VERSION,
+	"KERNEL_RELEASE":   OS_KERNEL_RELEASE,
+}
+
+// OSReleaseFieldToString is a map of os-release file fields
```
grantseltzer, line 88: nit: change to lowercase in godoc comment

helpers/osinfo.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// OSInfo object contains all OS relevant information
+//
+// OSRelease is relevant to examples such as:
+// 1) OSInfo.OSReleaseInfo[helpers.OS_KERNEL_RELEASE]) => will provide $(uname -r) string
+// 2) if OSInfo.OSRelease == helpers.UBUNTU => {} will allow to run code in specific distros
+//
+type OSInfo struct {
+	OSReleaseInfo     map[osReleaseField]string
```
grantseltzer, line 183: What does this field represent? How are consumers expected to use this map?

helpers/osinfo.go
rafaeldtinoco: Yep, made a mistake. Just realized that now. Thanks.

helpers/osinfo.go
rafaeldtinoco: Users are accessing it directly in the form of:

OSInfo.OSReleaseInfo[helpers.OS_KERNEL_RELEASE]

for example, instead of using a function to get the string. We could also do:

OSInfo.GetField(helpers.OS_KERNEL_RELEASE)

or any other similar form. We had mixed modes over libbpfgo and tracee, I've spoken to Itay today and we agreed that making internal API structures private was the best way to move forward. 

Should I change this ?

rafaeldtinoco:
Tests after this last push:

By using env variables pointing to default files:

```
$ sudo LIBBPFGO_KCONFIG_FILE=/boot/config-$(uname -r) LIBBPFGO_OSRELEASE_FILE=/etc/os-release ./dist/tracee-ebpf --debug --trace uid=1000 --trace pid=new --trace event=execve --trace event=cap_capable --output option:detect-syscall
OSInfo: VERSION_CODENAME: hirsute
OSInfo: KERNEL_RELEASE: 5.11.0-31-generic
OSInfo: VERSION: "21.04 (Hirsute Hippo)"
OSInfo: ID: ubuntu
OSInfo: ID_LIKE: debian
OSInfo: PRETTY_NAME: "Ubuntu 21.04"
OSInfo: VERSION_ID: "21.04"
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
```

By using env variables with files that does not exist. If tracee can't find /etc/os-release, or even /boot/config-$(uname -r) files, it should continue, as stated by Yaniv in other PRs.

```
$ sudo LIBBPFGO_KCONFIG_FILE=/config-$(uname -r).orig LIBBPFGO_OSRELEASE_FILE=/os-release.orig ./dist/tracee-ebpf --debug --trace uid=1000 --trace pid=new --trace event=execve --trace event=cap_capable --output option:detect-syscall
KConfig: warning: could not check enabled kconfig features
(could not open LIBBPFGO_KCONFIG_FILE /config-5.14.0-rc5+.orig)
OSInfo: KERNEL_RELEASE: 5.14.0-rc5+
OSInfo: warning: os-release file could be found
(could not open LIBBPFGO_OSRELEASE_FILE /os-release.orig)
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
```

By pointing to a confi.gz env file in archlinux (containing /proc/config.gz):

```
$ sudo LIBBPFGO_KCONFIG_FILE=/proc/config.gz LIBBPFGO_OSRELEASE_FILE=/etc/os-release.orig ./dist/tracee-ebpf --debug --trace uid=1000 --trace pid=new --trace event=execve --trace event=cap_capable --outputoption:detect-syscall
OSInfo: KERNEL_RELEASE: 5.13.13-arch1-1
OSInfo: warning: os-release file could be found
(could not open LIBBPFGO_OSRELEASE_FILE /etc/os-release.orig)
BTF: bpfenv = false, btfenv = false, vmlinux = true
BPF: using embedded BPF object
unpacked CO:RE bpf object file into memory
TIME             UID    COMM             PID     TID     RET              EVENT                ARGS
```

grantseltzer (approved):
LGTM :-)

PR #69, opened 2021-09-02
Add BPFLink.Destroy

derekparker:
There is currently no way to remove a BPF probe link with this library
so this patch adds support for the `bpf_link__destroy` function.


grantseltzer (approved):
LGTM! Thanks @derekparker !

PR #70, opened 2021-09-03
helpers/kernel_config.go: allow kconfig files logic

rafaeldtinoco:
Closes: #61

- fix nil returns from InitKernelConfig
- make kConfigFilePath private, accessible via GetKernelConfigFilePath()


rafaeldtinoco:
@yanivagman this is the last (hopefully) change to libbpfgo before the full "no os-release / no kconfig" tracee support (sending the PR shortly).

rafaeldtinoco:
Tests made and shown at: https://github.com/aquasecurity/libbpfgo/issues/61#issuecomment-912260355

rafaeldtinoco:
After this one is reviewed and merged, the following PR:

https://github.com/aquasecurity/tracee/pull/981

should also be  reviewed, have its libbpfgo version dependency bumped and merged.

helpers/kernel_config.go
```diff
@@ -88,6 +89,12 @@ const (
 	CONFIG_BPF_PRELOAD_UMD
 )
 
+// Constants declared here are used in the eBPF objects
+
+const (
+	CONFIG_ARCH_HAS_SYSCALL_WRAPPER KernelConfigOption = iota + 1000
```
yanivagman, line 95: Why do we need this constant?

rafaeldtinoco: To get it's value from kconfig file just as any other constant, in the same logic. Check Tracee PR to see I'm reading this kconfig variable value using this constant. We can extend this and I wanted to differentiate for the consts we use to check supportability.

yanivagman: I see.
So, anytime we (or any other user of libbpfgo) will want to add some new config options (and there might be many of these), we will have to update this KernelConfigOption enum? I don't think that it will scale well...

When thinking about it, the KernelConfigOption values have no meaning, just an enumeration, right? So my question is, why do we even need this enum to be defined as int32? Why not defining it as a string? Anyways we convert it from string and to string, but do we really need these conversions? The natural type for this value would be string, am I wrong? And if we only need to make these conversions to int just so we can use it as a key to a bpf map, then I think that the responsibility of making such conversions should be of the user of libbpfgo (e.g. tracee), and not of libbpfgo itself. WDYT?

rafaeldtinoco: Yes makes sense. The user (Tracee) is the one to be in charge of an array of map[uint32]strings that shall be used as source for the kconfig map - or something similar - so we don't have to specify inside the library but yet have it defined to kconfig_map.. let me work on that..

rafaeldtinoco: @yanivagman done. Please check the new version together with https://github.com/aquasecurity/tracee/pull/981

helpers/kernel_config.go
```diff
@@ -283,8 +307,17 @@ func (k *KernelConfig) readConfigFromScanner(reader io.Reader) {
 }
 
 // GetValue will return a KernelConfigOptionValue for a given KernelConfigOption when this is a BUILTIN or a MODULE
-func (k *KernelConfig) GetValue(option KernelConfigOption) (KernelConfigOptionValue, error) {
-	value, ok := k.configs[option].(KernelConfigOptionValue)
+func (k *KernelConfig) GetValue(option interface{}) (KernelConfigOptionValue, error) {
```
grantseltzer, line 310: Why does this take an option as interface{} ?

rafaeldtinoco: because I have to feed a uint32 from tracee as Im not adding KernelConfigOption (check consts.go in tracee with newly added uint32 const).

rafaeldtinoco: Okay, it does not have to.. the const can be declared as helper.KernelConfigOption as well I suppose.

helpers/kernel_config.go
```diff
@@ -174,53 +175,71 @@ var KernelConfigKeyIDToString = map[KernelConfigOption]string{
...
+	return &config, err
+}
+
+// GetKernelConfigFilePath gives the kconfig file chosen by InitKernelConfig during initialization
+func (k *KernelConfig) GetKernelConfigFilePath() string {
+	return k.kConfigFilePath
+}
+
+// AddCustomKernelConfigs allows user to extend list of possible existing kconfigs to be parsed from kConfigFilePath
+func (k *KernelConfig) AddCustomKernelConfigs(values map[KernelConfigOption]string) error {
```
grantseltzer, line 235: I'd prefer this method to just take a key and value, let the user iterate over their map and call it. 

Also, what exactly is the usecase for it? And should we have validation for the keys being real config options? Is it to edit existing keys or add new ones? If it's to add new ones we should validate that it doesn't already exist.

rafaeldtinoco: There is no need for validation as the values can indeed be Undefined because they don't exist in running kconfig file but they could exist in some others. We would have to have an updated list of possible config options for new kernels as well and it does no make much sense. 

User has to inform option and if it exists it's value will be defined, if not.. undefined.

rafaeldtinoco: About use case and iteration, could you check my pull request to Tracee as a consumer to this and see if you still think that is the best way ? Thanks

rafaeldtinoco: Iteration with single items would make it equal to initial logic of pushing needed config options one by one.. maybe it's the best way forward indeed .. to keep everything similar.

grantseltzer (approved):
LGTM!

helpers/kernel_config.go
```diff
@@ -174,53 +175,69 @@ var KernelConfigKeyIDToString = map[KernelConfigOption]string{
...
+// GetKernelConfigFilePath gives the kconfig file chosen by InitKernelConfig during initialization
+func (k *KernelConfig) GetKernelConfigFilePath() string {
+	return k.kConfigFilePath
+}
+
+// AddCustomKernelConfigs allows user to extend list of possible existing kconfigs to be parsed from kConfigFilePath
+func (k *KernelConfig) AddCustomKernelConfigs(key KernelConfigOption, value string) error {
+	// extend initial list of kconfig options: add other possible existing ones
+	KernelConfigKeyIDToString[key] = value
+	KernelConfigKeyStringToID[value] = key
```
yanivagman, line 238: What if the given key already exists (e.g. any number between 1-30 will collide with the already existing keys. Do we want to allow this? Warn? Error?

Also, it is not documented which number the user should choose for the key (I saw that in Tracee you used 1000 as the starting point arbitrarily).

Using strings as keys could have resolved this (as I suggested in a previous comment). If that is problematic to change, here is another suggestion:
1. Define a new constant for the starting point of custom keys (e.g. CUSTOM_CONFIG_START = 1000)
2. Enforce custom kernel configs to be bigger than this value and return error if not

WDYT?

helpers/kernel_config.go
```diff
@@ -174,53 +175,69 @@ var KernelConfigKeyIDToString = map[KernelConfigOption]string{
...
 
-	return &config, nil
+	return k.initKernelConfig(k.kConfigFilePath)
```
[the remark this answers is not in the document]
rafaeldtinoco: This is that case of .. after reviews it got weird at the end. This function accepted a map[KernelConfigOption][string] to extend all supported kernel configs with a bunch of added other options. Now, accepting a single argument.. re-reading all the file every time there is an update is weird indeed.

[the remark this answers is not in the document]
rafaeldtinoco: Ah ok.. for this one, for the record.. we're not saving ALL existing kconfig values into existing object's memory, only the ones we have knowledge of (this was already done). So, whenever we add new KernelConfigOption's, we need to re-read the kconfig file to set its value. 

If I keep value of all existing kconfig options in the 1st parse, then there is not even a need for AddCustomKernelConfigs() function. Since I prefer not to alter this working (and tested) logic currently, I'm adding a TODO here to improve that so we don't have to parse the file multiple times. It no big deal as the file will be cached (or in procfs), but, still.. not nice.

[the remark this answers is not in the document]
rafaeldtinoco: addressed at: https://github.com/aquasecurity/libbpfgo/pull/73 (not ideal, but better than reloading each time a new custom option is added).

helpers/kernel_config.go
rafaeldtinoco: Addressed at: https://github.com/aquasecurity/libbpfgo/pull/73

About the strings, and not following that path from your previous comment, in some review I got another feedback to close the options (instead of having them opened as strings to be included). Nevertheless, I think it is working now and has better error handling than before the new PR mentioned. 

We can add an issue to improve this if it proves to be too bad by design. The way it is right now we can add the kconfig options and use them in tracee.bpf.c.

rafaeldtinoco:
FTR, I've go some minor continuation of this at: https://github.com/aquasecurity/libbpfgo/pull/73

PR #83, opened 2021-09-27
Fix NewModuleFromBuffer

derekparker:
NewModuleFromBuffer was calling NewModuleFromFileArgs instead of
NewModuleFromBufferArgs which ended up causing the following error to
start coming through:

libbpf: elf: failed to open : No such file or directory

This patch fixes it by calling the correct function.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -197,7 +197,6 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


rafaeldtinoco (approved):
Thanks for having spotting that. Bad `yy + p` when I last touched this.

Accepting it.

PR #84, opened 2021-10-26
Add `no_bpf` build tag to disable the dependencies

saschagrunert:
To be able to use libbpfgo optionally in projects we require a build tag
to disable the third party cgo dependencies. This can be now done as
opt-out by using the `no_bpf` build tag.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -64,6 +64,15 @@ $ make selftest-static-run => will build & run all static selftests
 > Note 01: dynamic builds need your OS to have a *recent enough* libbpf package (and its headers) installed. Sometimes, recent features might require the use of backported OS packages in order for your OS to contain latest *libbpf* features (sometimes required by libbpfgo).
 > Note 02: static builds need `git submodule init` first. Make sure to sync the *libbpf* git submodule before trying to statically compile or test the *libbpfgo* repository.
 
+### Build tags
+
+If a project wants to make the libbpfgo dependencies (libbpf, libz, libelf)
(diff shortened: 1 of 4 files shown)


saschagrunert:
@rafaeldtinoco PTAL

grantseltzer:
Hi @saschagrunert I'm sorry about the delay in review, why might a project want to make an imported dependency optional? What's the use case for this?

saschagrunert:
> Hi @saschagrunert I'm sorry about the delay in review, why might a project want to make an imported dependency optional? What's the use case for this?

Hey, for example in our security profiles operator we have a bpf based seccomp profile recorder as PoC integrated right now (still optimizing it). The component is configurable and can be also optionally compiled out for build environments not featuring libbpf (0.5.0 to be exact). 

Do you think this nothing we would like to support? Feel free to close if so and if it's out of scope of this project.

yanivagman:
Not sure I understand - Why can't this directive (`// +build !no_bpf`) be used by the consumer of libbpfgo when he want to disable the third party cgo dependencies?

saschagrunert:
> Not sure I understand - Why can't this directive (`// +build !no_bpf`) be used by the consumer of libbpfgo when he wants to disable third party cgo dependencies?

Ah yeah I think I had an issue in my initial implementation, testing with the build tag from the root project should work as well. Thank you again folks, closing this one now.

grantseltzer:
@saschagrunert Thanks for your contribution/discussion nonetheless!

PR #87, opened 2021-11-05
helpers/osinfo: host architecture detection

rafaeldtinoco:
osinfo needs to detect the host architecture in order to support the embedded BTF files feature being added to tracee.

### DIFF
--- a/helpers/common.go
+++ b/helpers/common.go
@@ -38,6 +38,28 @@ func UnameRelease() (string, error) {
 	return ver, nil
 }
 
+// UnameMachine gets the version string of host's architecture
+func UnameMachine() (string, error) {
+	var uname syscall.Utsname
+	if err := syscall.Uname(&uname); err != nil {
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
As soon as this is merged I'll do the PR for BTFHub support for tracee.

helpers/common.go
```diff
@@ -38,6 +38,28 @@ func UnameRelease() (string, error) {
...
+	var buf [65]byte
+	for i, b := range uname.Machine {
+		buf[i] = byte(b)
+	}
+
+	arch := string(buf[:])
+	arch = strings.Trim(arch, "\x00")
+
+	if strings.Contains(arch, "aarch64") {
+		arch = "x86_64"
```
yanivagman, line 57: shouldn't this be "arm64"?

rafaeldtinoco: Yes, fixing it.

yanivagman (approved):

Issue #88, opened 2021-11-08
Create helpers for checking raw arguments


grantseltzer:
In [tracee PR #1123](https://github.com/aquasecurity/tracee/pull/1123) parsing of syscall arguments into strings can be turned on with an argument to the output flag. We should provide helpers for checking particular values in the raw 

For example, the clone 'flag' argument has [a bunch of options](https://linux.die.net/man/2/clone), we could provide helpers like `bool HasClearTID(flags)` to detect if `flags` contains `CLONE_CHILD_CLEARTID`

PR #89, opened 2021-11-11
Parsing syscall flag helper and constants

grantseltzer:
#88


helpers/argumentParsers.go
```diff
@@ -107,61 +107,61 @@ func ParseOpenFlags(flags uint32) string {
 
 	// access mode
 	switch {
-	case flags&01 == 01:
+	case eventArgumentContainsOption(uint64(flags), O_WRONLY):
```
itaysk, line 110: Since we are now execting external users to use these consts (since Tracee is emitting raw values), maybe it would be better to strongly-type them and to implement the `stringer` interface on the types: https://pkg.go.dev/fmt#Stringer
This way we (and others) can just `fmt.Println(O_WRONLY)` instead of the developer having to check the const value and then type the const name as string, like seen in this code block.
Possibly we can use the stringer utility to automate this: https://pkg.go.dev/golang.org/x/tools/cmd/stringer

itaysk: In other words, I'm saying these identifier-to-string mapping could be moved to where we define the consts, instead of where we print them, and by that they would become reusable

helpers/syscall_args.go
```diff
@@ -0,0 +1,83 @@
...
+// EventArgumentContainsOption checks whether the argument (rawArgument)
+// contains the 'option' such as with flags passed to the clone flag.
+// Typically linux syscalls have multiple options specified in a single
+// argument via bitmasks, which this function checks for.
+// It is meant to be used with the constants redefined in this package.
+func EventArgumentContainsOption(option, rawArgument uint64) bool {
+	return eventArgumentContainsOption(option, rawArgument)
+}
+
+func eventArgumentContainsOption(option, rawArgument uint64) bool {
```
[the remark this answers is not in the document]
grantseltzer: It was for more flexibility in case we needed to change something without breaking API. In general I don't like having exported API functions call other exported API functions directly. But this doesn't apply after my latest rework.

grantseltzer:
@rafaeldtinoco @mtcherni95 can I ask for either of your review here?

mtcherni95 (approved):
LGTM, @rafaeldtinoco what do you think?

grantseltzer:
@yanivagman This PR makes changes to the parse helpers. I'd like to merge these before my patch for #1029 that I have ready waiting in the wings. Can you take a look at this?

helpers/argumentParsers.go
```diff
@@ -2,622 +2,1374 @@ package helpers
...
-// https://elixir.bootlin.com/linux/v5.5.3/source/include/uapi/asm-generic/mman-common.h#L10
-func ParseMemProt(prot uint32) string {
+type CloneFlagArgument struct {
+	rawValue    uint64
+	stringValue string
+}
+
+func (c CloneFlagArgument) Value() uint64  { return c.rawValue }
+func (c CloneFlagArgument) String() string { return c.stringValue }
+
```
yanivagman, line 359: It might be clearer if we put the CloneFlagArgument values defined above here so it is grouped in the same place.
Same for all other types.
WDYT?

helpers/argumentParsers.go
```diff
@@ -2,622 +2,1374 @@ package helpers
...
+		f = append(f, CLONE_NEWPID.String())
+	}
+	if OptionAreContainedInArgument(rawValue, CLONE_NEWNET) {
+		f = append(f, CLONE_NEWNET.String())
+	}
+	if OptionAreContainedInArgument(rawValue, CLONE_IO) {
+		f = append(f, CLONE_IO.String())
+	}
+	if len(f) == 0 {
+		return CloneFlagArgument{}, fmt.Errorf("no valid clone flag values present in raw value: 0x%x", rawValue)
```
yanivagman, line 435: It might be that no flags were given as an argument (rawValue is 0), why do we want to return error in that case? Shouldn't we just return an empty string?
Same for all other options below.

helpers/argumentParsers.go
```diff
@@ -2,622 +2,1374 @@ package helpers
...
+func (mode InodeModeArgument) Value() uint64  { return mode.rawValue }
+func (mode InodeModeArgument) String() string { return mode.stringValue }
+
+func ParseInodeMode(rawValue uint64) (InodeModeArgument, error) {
+	var f []string
+
+	// File Type
+	switch {
+	case OptionAreContainedInArgument(rawValue, S_IFSOCK):
+		f = append(f, S_IFSOCK.String())
```
yanivagman, line 1258: Here you return a string, while in other "Parse" functions you return the SystemFunctionArgument type (value+string). I think we should be consistent for all of the "Parse" functions.

helpers/argumentParsers.go
```diff
@@ -2,622 +2,1374 @@ package helpers
...
+	PTRACE_GETREGS              PtraceRequestArgument = 12
+	PTRACE_SETREGS              PtraceRequestArgument = 13
+	PTRACE_GETFPREGS            PtraceRequestArgument = 14
+	PTRACE_SETFPREGS            PtraceRequestArgument = 15
+	PTRACE_ATTACH               PtraceRequestArgument = 16
+	PTRACE_DETACH               PtraceRequestArgument = 17
+	PTRACE_GETFPXREGS           PtraceRequestArgument = 18
+	PTRACE_SETFPXREGS           PtraceRequestArgument = 19
+	PTRACE_SYSCALL              PtraceRequestArgument = 24
+	PTRACE_SETOPTIONS           PtraceRequestArgument = 0x4200
```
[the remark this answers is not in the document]
grantseltzer: I do agree, but this is how it is in [ptrace.h](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/ptrace.h#L25). I think it's more important to maintain the convention of copying the header than hex/decimal.

helpers/argumentParsers.go
grantseltzer: `ParseInodeMode` returns an InodeModeArgument but I did see the other parse function which returned a string, did you comment on the wrong line here?

helpers/argumentParsers.go
grantseltzer: Sure, I can organize it that way

helpers/argumentParsers.go
grantseltzer: This is a good point, perhaps it should be 

`if len(f) == 0 && rawValue != 0 {`

Which would mean that it isn't an empty argument, but it isn't any valid one. WDYT?

yanivagman: Another option is to check at the beginning of the function if `rawValue == 0 ` and return an empty string.
Then you can keep this test here

helpers/argumentParsers.go
yanivagman: Yes, I thought you return the string value, but I now see it returns InodeModeArgument

yanivagman (approved):
LGTM

PR #94, opened 2021-11-19
Add helper to check if ftrace is enabled

grantseltzer:

### DIFF
--- a/helpers/osinfo.go
+++ b/helpers/osinfo.go
@@ -2,6 +2,8 @@ package helpers
 
 import (
 	"bufio"
+	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -218,4 +220,16 @@ func (btfi *OSInfo) discoverOSDistro() error {
 	}
 
 	return nil
+}
+
+func FtraceEnabled() (bool, error) {
+	b, err := os.ReadFile("/proc/sys/kernel/ftrace_enabled")
+	if err != nil {
+		return false, fmt.Errorf("could not read from ftrace_enabled file: %s", err.Error())
+	}
+	b = bytes.TrimSpace(b)
+	if len(b) != 1 {
+		return false, errors.New("malformed ftrace_enabled file")
+	}
+	return b[0] == '1', nil
 }


helpers/osinfo.go
```diff
@@ -219,3 +220,11 @@ func (btfi *OSInfo) discoverOSDistro() error {
...
 	return nil
 }
+
+func FtraceEnabled() bool {
+	b, err := ioutil.ReadFile("/proc/sys/kernel/ftrace_enabled")
+	if err != nil || len(b) != 1 || b[0] == '0' {
+		return false
+	}
+	return true
+}
```
rafaeldtinoco, line 230: $ echo 1 | sudo tee /proc/sys/kernel/ftrace_enabled
1

$ go run temp.go
false

$ echo 0 | sudo tee /proc/sys/kernel/ftrace_enabled
0

$ go run temp.go
false

rafaeldtinoco: Something is wrong, no ?

```
[rafaeldtinoco@fujitsu:~]$ cat temp.go
package main

import (
	"fmt"
	"io/ioutil"
)

func FtraceEnabled() bool {
	b, err := ioutil.ReadFile("/proc/sys/kernel/ftrace_enabled")
	c := string(b)
	if c[0] == '0' {
		fmt.Println(false)
	} else if c[0] == '1' {
		fmt.Println(true)
	}

	if err != nil || len(b) != 1 || b[0] == '0' {
		return false
	}
	return true
}

func main() {
	fmt.Println(FtraceEnabled())
}
```
```
[rafaeldtinoco@fujitsu:~]$ go run temp.go
true
false
```

grantseltzer: Ah woops, there's an EOL appended when reading this way. I'll fix it.

grantseltzer: Ok it appears to always append the EOL so I fixed it accordingly.

rafaeldtinoco (approved):
LGTM

mtcherni95 (commented):
why not just use TrimSpace and compare it string? don't we also want to report the error instead of just a boolean?

rafaeldtinoco:
> why not just use TrimSpace and compare it? don't we also want to report the error instead of just a boolean?

I thought you had agreed already on this helper (when I saw your other PR). Whatever you both prefer I'm good, its a simple change.

mtcherni95:
> 

in the other pr I implemented what I suggested above - we can use return both bool and error IMO.

rafaeldtinoco:
> in the other pr I implemented what I suggested above - we can use return both bool and error IMO.

@mtcherni95 and @yanivagman , I dont recall what we've agreed in last weeks discussion. Was it to be placed in libbpfgo or tracee ? I recall it *could* have been agreed to be added to tracee because libbpf did not have that check, but I could be wrong.

mtcherni95:
> > in the other pr I implemented what I suggested above - we can use return both bool and error IMO.
> 
> @mtcherni95 and @yanivagman , I dont recall what we've agreed in last weeks discussion. Was it to be placed in libbpfgo or tracee ? I recall it _could_ have been agreed to be added to tracee because libbpf did not have that check, but I could be wrong.

You are right - we agreed it would have been in tracee. However, after some further thinking it is better to have it in libbpgo and tracee will use the new API of FtraceEnabled

grantseltzer:
> why not just use TrimSpace and compare it? don't we also want to report the error instead of just a boolean?

Agreed, changed accordingly

helpers/osinfo.go
```diff
@@ -220,10 +221,12 @@ func (btfi *OSInfo) discoverOSDistro() error {
 	return nil
 }
 
-func FtraceEnabled() bool {
+func FtraceEnabled() (bool, error) {
 	b, err := os.ReadFile("/proc/sys/kernel/ftrace_enabled")
-	if err != nil || len(b) != 2 || b[0] == '0' {
-		return false
+	b = bytes.TrimSpace(b)
```
rafaeldtinoco, line 226: this should be after the error check, no ?

rafaeldtinoco (commented):
would you mind squashing the changes into a single commit and addressing issue from my comment ? this way we can move on and merge this. thanks!

helpers/osinfo.go
grantseltzer: My mistake. Also added a check for length to avoid an awkward go panic

rafaeldtinoco (approved):
LGTM:

```
$ echo 0 | sudo tee /proc/sys/kernel/ftrace_enabled
0
$ go run temp.go
false <nil>
$ echo 1 | sudo tee /proc/sys/kernel/ftrace_enabled
1
$ go run temp.go
true <nil>
```
and simulating an error
```
$ go run temp.go
false could not read from ftrace_enabled file: open /proc/sys/kernel/ftrace_enabled: no such file or directory
```

PR #97, opened 2021-12-13
libbpfgo: Add map batch operations

derekparker:
This patch adds a few batch operations for maps including:

Lookup, LookupAndDelete, Update and Delete.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,13 +76,13 @@ libbpfgo-static: $(VMLINUXH) | $(LIBBPF_OBJ)
 
 libbpfgo-static-test: libbpfgo-test-bpf-static
 	sudo env PATH=$(PATH) \
-	CC=$(CLANG) \
-	CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
-	CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
-	GOOS=linux GOARCH=$(ARCH) \
-	go test \
-	-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
-	.
+		CC=$(CLANG) \
+		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
+		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
+		GOOS=linux GOARCH=$(ARCH) \
+		go test \
+		-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
+		.
 
 # vmlinux header file
 
--- a/Makefile
+++ b/Makefile
@@ -75,13 +75,14 @@ libbpfgo-static: $(VMLINUXH) | $(LIBBPF_OBJ)
 		.
 
 libbpfgo-static-test: libbpfgo-test-bpf-static
+	sudo env PATH=$(PATH) \
 	CC=$(CLANG) \
-		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
-		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
-		GOOS=linux GOARCH=$(ARCH) \
-		sudo -E -- go test \
-		-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
-		.
+	CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
+	CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
+	GOOS=linux GOARCH=$(ARCH) \
+	go test \
+	-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
+	.
 
 # vmlinux header file
 
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ libbpfgo-static-test: libbpfgo-test-bpf-static
 		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
 		GOOS=linux GOARCH=$(ARCH) \
 		go test \
-		-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
+		-v -tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
 		.
 
 # vmlinux header file
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -600,6 +600,10 @@ func bpfMapBatchOptsToC(batchOpts *BPFMapBatchOpts) *C.struct_bpf_map_batch_opts
 // GetValueBatch allows for batch lookups of multiple keys.
 // The first argument is a pointer to an array or slice of keys which will be populated with the keys returned from this operation.
 // It returns the associated values as a slice of slices of bytes.
+// This API allows for batch lookups of multiple keys, potentially in steps over multiple iterations. For example,
+// you provide the last key seen (or nil) for the startKey, and the first key to start the next iteration with in nextKey.
+// Once the first iteration is complete you can provide the last key seen in the previous iteration as the startKey for the next iteration
+// and for forth until nextKey is nil.
 func (b *BPFMap) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
 	var (
 		values    = make([]byte, b.ValueSize()*int(count))
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -598,7 +598,7 @@ func bpfMapBatchOptsToC(batchOpts *BPFMapBatchOpts) *C.struct_bpf_map_batch_opts
 }
 
 // GetValueBatch allows for batch lookups of multiple keys.
-// The first argument is a pointer to an array of keys which will be populated with the keys returned from this operation.
+// The first argument is a pointer to an array or slice of keys which will be populated with the keys returned from this operation.
 // It returns the associated values as a slice of slices of bytes.
 func (b *BPFMap) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
 	var (
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
 	return value, nil
 }
 
+// BPFMapBatchOpts mirrors the C structure bpf_map_batch_opts.
+type BPFMapBatchOpts struct {
+	Sz        uint64
+	ElemFlags uint64
+	Flags     uint64
+}
+
+func bpfMapBatchOptsToC(batchOpts *BPFMapBatchOpts) *C.struct_bpf_map_batch_opts {
+	if batchOpts == nil {
+		return nil
+	}
+	opts := C.struct_bpf_map_batch_opts{}
+	opts.sz = C.ulong(batchOpts.Sz)
+	opts.elem_flags = C.ulonglong(batchOpts.ElemFlags)
+	opts.flags = C.ulonglong(batchOpts.Flags)
+
+	return &opts
+}
+
+// GetValueBatch allows for batch lookups of multiple keys.
+// The first argument is a pointer to an array of keys which will be populated with the keys returned from this operation.
+// It returns the associated values as a slice of slices of bytes.
+func (b *BPFMap) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
+	var (
+		values    = make([]byte, b.ValueSize()*int(count))
+		valuesPtr = unsafe.Pointer(&values[0])
+		countC    = C.uint(count)
+	)
+
+	opts := &BPFMapBatchOpts{
+		Sz:        uint64(unsafe.Sizeof(BPFMapBatchOpts{})),
+		ElemFlags: C.BPF_ANY,
+		Flags:     C.BPF_ANY,
+	}
+
+	errC := C.bpf_map_lookup_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
+	if errC != 0 {
+		return nil, fmt.Errorf("failed to batch lookup values %v in map %s: %d", keys, b.name, errC)
+	}
+
+	parsedVals := collectBatchValues(values, count, b.ValueSize())
+
+	return parsedVals, nil
+}
+
+// GetValueAndDeleteBatch allows for batch lookups of multiple keys and deletes those keys.
+func (b *BPFMap) GetValueAndDeleteBatch(keys, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
+	var (
+		values    = make([]byte, b.ValueSize()*int(count))
+		valuesPtr = unsafe.Pointer(&values[0])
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -75,13 +75,14 @@ libbpfgo-static: $(VMLINUXH) | $(LIBBPF_OBJ)
...
 libbpfgo-static-test: libbpfgo-test-bpf-static
+	sudo env PATH=$(PATH) \
 	CC=$(CLANG) \
-		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
-		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
-		GOOS=linux GOARCH=$(ARCH) \
-		sudo -E -- go test \
-		-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
-		.
+	CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
```
rafaeldtinoco, line 80: could you, pls, keep the indentation here as it was ? I'm going to "fix" the Makefile style soon, not only for this line but for other lines as well.

derekparker: Ack, no problem!

rafaeldtinoco (dismissed):
LGTM. We should open an issue to transform the "selftest/build" into a proper selftest/XXX directory. @grantseltzer WDYT ?

Besides, I think we should "unify" the Makefiles into a single Makefile at the root dir (just like we recently did to tracee). 

Nevertheless, I think we should accept this contribution, which was great.

@grantseltzer I'll let you squash & merge this if you agree.

libbpfgo.go
```diff
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	opts := &BPFMapBatchOpts{
+		Sz:        uint64(unsafe.Sizeof(BPFMapBatchOpts{})),
+		ElemFlags: C.BPF_ANY,
+		Flags:     C.BPF_ANY,
+	}
+
+	errC := C.bpf_map_lookup_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
+	if errC != 0 {
+		return nil, fmt.Errorf("failed to batch lookup values %v in map %s: %d", keys, b.name, errC)
+	}
```
grantseltzer, line 619: errno needs to be checked here, you can use syscall.Errno like elsewhere in libbpfgo.

libbpfgo.go
```diff
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	opts := &BPFMapBatchOpts{
+		Sz:        uint64(unsafe.Sizeof(BPFMapBatchOpts{})),
+		ElemFlags: C.BPF_ANY,
+		Flags:     C.BPF_ANY,
+	}
+
+	errC := C.bpf_map_lookup_and_delete_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
+	if errC != 0 {
+		return nil, fmt.Errorf("failed to batch lookup and delete values %v in map %s", keys, b.name)
+	}
```
grantseltzer, line 643: errno needs to be checked here, you can use syscall.Errno like elsewhere in libbpfgo.

libbpfgo.go
```diff
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	countC := C.uint(count)
+	opts := &BPFMapBatchOpts{
+		Sz:        uint64(unsafe.Sizeof(BPFMapBatchOpts{})),
+		ElemFlags: C.BPF_ANY,
+		Flags:     C.BPF_ANY,
+	}
+	errC := C.bpf_map_delete_batch(b.fd, keys, &countC, bpfMapBatchOptsToC(opts))
+	if errC != 0 {
+		return fmt.Errorf("failed to get lookup key %d from map %s", keys, b.name)
+	}
```
grantseltzer, line 687: errno needs to be checked here, you can use syscall.Errno like elsewhere in libbpfgo.

libbpfgo.go
```diff
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	opts := BPFMapBatchOpts{
+		Sz:        uint64(unsafe.Sizeof(BPFMapBatchOpts{})),
+		ElemFlags: C.BPF_ANY,
+		Flags:     C.BPF_ANY,
+	}
+	errC := C.bpf_map_update_batch(b.fd, keys, values, &countC, bpfMapBatchOptsToC(&opts))
+	if errC != 0 {
+		return fmt.Errorf("failed to update map %s: %v", b.name, errC)
+	}
+	return nil
```
grantseltzer, line 672: errno needs to be checked here, you can use syscall.Errno like elsewhere in libbpfgo.

Makefile
```diff
@@ -75,11 +75,12 @@ libbpfgo-static: $(VMLINUXH) | $(LIBBPF_OBJ)
...
 libbpfgo-static-test: libbpfgo-test-bpf-static
-	CC=$(CLANG) \
+	sudo env PATH=$(PATH) \
+		CC=$(CLANG) \
 		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
 		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
 		GOOS=linux GOARCH=$(ARCH) \
-		sudo -E -- go test \
+		go test \
 		-tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
```
grantseltzer, line 84: can you add `-v` in here while you're at it?

grantseltzer (changes requested):
Thanks so much for this contribution Derek! 

In addition to my inline comments, can I ask that you move your tests into `selftests`? It shouldn't make much of a change, just copy the directory structure of one of the other selftests.

I'm not entirely sure why we still have the test that's in `libbpfgo_test.go` (Test_LoadAndAttach), that file should contain tests that don't shell out, things like helper functions and whatnot. I'm going to create an issue to track removing/replacing that one.

libbpfgo.go
```diff
@@ -578,6 +578,116 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	opts.sz = C.ulong(batchOpts.Sz)
+	opts.elem_flags = C.ulonglong(batchOpts.ElemFlags)
+	opts.flags = C.ulonglong(batchOpts.Flags)
+
+	return &opts
+}
+
+// GetValueBatch allows for batch lookups of multiple keys.
+// The first argument is a pointer to an array of keys which will be populated with the keys returned from this operation.
+// It returns the associated values as a slice of slices of bytes.
```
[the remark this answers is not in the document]
derekparker: That SGTM, will update.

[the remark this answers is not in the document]
derekparker: Actually, the way I would implement this would be to return `map[[]byte][]byte` because we don't know what type the keys will be (or size). The problem with this is a `[]byte` is not comparable so cannot be used as a key. I suppose we could use a `map[interface{}][]byte` and that should work (haven't tested it) but feels kind of weird to me, and adds a type assertion to the callers code.

I'm happy to implement the latter if that's what you'd prefer, however this interface matches what happens in libbpf (although admittedly it's not very Go-like).

As shown in the test it's fairly straightforward still to iterate through keys / values using a single iterator var to keep track of where you are.

libbpfgo.go
derekparker: Apologies, I followed the example of the other BPFMap methods (such as `Pin`, `UnPin`, etc...) which don't do any extra checking on the returned error than what I've done. Do you have another specific example or pattern you'd like me to follow or extra checks to perform?

Makefile
derekparker: Done.

derekparker:
Hey @grantseltzer thanks for the feedback!

I've addressed most of your review except for that which I've commented on. Also I've moved the test over to it's own directory in `selftest/map-batch` and removed the corresponding tests in `libbpfgo_test.go`.

libbpfgo.go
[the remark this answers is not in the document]
grantseltzer: Good points, best to leave it as is.

libbpfgo.go
grantseltzer: I had an answer ready to this but then after carefully examining libbpf code I got confused. I asked on the mailing list how functions which use `libbpf_err` vs `libbpf_err_errno` are supposed to be checked, let's wait for a response from Andrii before proceededing:

https://lore.kernel.org/bpf/CAO658oUqd5=B3zkDhm2jVQxG+vEf=2CE7WimXHqgcH+m0P=k_Q@mail.gmail.com/T/#u

I'm sure there's functions in libbpfgo that aren't correct.

grantseltzer: @yanivagman You probably have a good idea

libbpfgo.go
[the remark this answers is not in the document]
derekparker: Sounds good!

libbpfgo.go
derekparker: Ack, sounds good to me!

derekparker: If I understand correctly it looks like Andrii responded saying that the returned value can be used directly without checking errno unless it's an API which returns a pointer, in which case errno must be checked.

Are you reading the response the same?

grantseltzer: Yes I read it the same, which means the way you're doing it is correct. Though perhaps we can extract the error code to determine what the issue was. I think that's what using the `syscall.Errno` type does, but I will investigate in the morning.

selftest/map-batch/main.go
```diff
@@ -0,0 +1,128 @@
...
+	batchKeys := make([]uint32, 3)
+	nextKey := uint32(0)
+	prevKey := unsafe.Pointer(nil)
+	step := len(batchKeys)
+	for i := 0; i < 2; i++ {
+		if i > 0 {
+			// We're on step 2, so test the batch lookup by specifying the
+			// key to start from.
+			prevKey = unsafe.Pointer(&nextKey)
+		}
```
grantseltzer, line 82: Can you explain more of what's going on here? Why is there two steps? I see the documentation for BPF_MAP_LOOKUP_BATCH in uapi/linux/bpf.h but I don't fully understand why it's designed this way.

selftest/map-batch/main.bpf.c
```diff
@@ -0,0 +1,43 @@
...
+    __type(value, struct value);
+    __uint(max_entries, 1<<24);
+} tester SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+    __uint(key_size, sizeof(u32));
+    __uint(value_size, sizeof(u32));
+} events SEC(".maps");
+
```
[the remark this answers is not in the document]
derekparker: Removed.

selftest/map-batch/main.bpf.c
```diff
@@ -0,0 +1,43 @@
+//+build ignore
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+
+struct value {
+    int x;
+    char y;
+};
```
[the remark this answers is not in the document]
derekparker: Done, forgot to do that when swapping over.

selftest/map-batch/main.bpf.c
```diff
@@ -0,0 +1,43 @@
...
+
+    int64_t secondKey = 42069420;
+    struct value *v2 = bpf_map_lookup_elem(&tester, &secondKey);
+    if (!v2) {
+        return 1;
+    }
+    bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, v2, sizeof(char)*3);
+
+    return 0;
+}
```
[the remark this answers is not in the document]
derekparker: Done.

selftest/map-batch/main.go
```diff
@@ -0,0 +1,128 @@
...
+		vals, err := testerMap.GetValueBatch(unsafe.Pointer(&batchKeys[0]), prevKey, unsafe.Pointer(&nextKey), uint32(step))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "step %d testerMap.LookupBatch failed: %v", i, err)
+			os.Exit(-1)
+		}
```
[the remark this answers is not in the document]
derekparker: Ack, that's my bad forgot to update when swapping over from the other test.

selftest/map-batch/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+	defer bpfModule.Close()
+
+	if err = resizeMap(bpfModule, "events", 8192); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
```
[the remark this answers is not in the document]
derekparker: Removed.

selftest/map-batch/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if err = m.Resize(size); err != nil {
+		return err
+	}
+
+	if actual := m.GetMaxEntries(); actual != size {
+		return fmt.Errorf("map resize failed, expected %v, actual %v", size, actual)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
derekparker: Removed.

selftest/map-batch/main.go
derekparker: Yeah, so the API is such that you can iterate over all the keys/values in batches. Essentially you provide a pointer to the last key seen, the next key to start with, and a buffer for keys/values.

That way if you have a map with max size of 1000 for example you can iterate over the entire thing in chunks instead of loading everything into memory. It's kind of a confusing API to be honest it took me a bit to wrap my head around it when implementing.

grantseltzer: Got it - that took some time to understand. Definitely worth documenting upstream in libbpf.

grantseltzer: After that this is good to go :+1:

derekparker: > Got it - that took some time to understand. Definitely worth documenting upstream in libbpf.

Yes! I literally had to read through the test code to understand how to use the API.

derekparker: Ok, let me know if the documentation I added is clear enough.

libbpfgo.go
```diff
@@ -578,6 +578,120 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
...
+	return &opts
+}
+
+// GetValueBatch allows for batch lookups of multiple keys.
+// The first argument is a pointer to an array or slice of keys which will be populated with the keys returned from this operation.
+// It returns the associated values as a slice of slices of bytes.
+// This API allows for batch lookups of multiple keys, potentially in steps over multiple iterations. For example,
+// you provide the last key seen (or nil) for the startKey, and the first key to start the next iteration with in nextKey.
+// Once the first iteration is complete you can provide the last key seen in the previous iteration as the startKey for the next iteration
+// and for forth until nextKey is nil.
```
grantseltzer, line 606:
```suggestion
// and repeat until nextKey is nil.
```

grantseltzer:
Made a small edit I didn't want to bother you with. This looks good to me! Thanks so much @derekparker !

grantseltzer (approved):

PR #100, opened 2021-12-16
Standardize error checking across libbpfgo

grantseltzer:
This handles correcting how we check errors in libbpfgo
in accordance with how libbpf recommends handling errors
leading into libbpf.

Particularly this follows the following rules:

- APIs that return an error code (int) should have
  error codes checked directly. The error codes
  correspond with error codes in the syscall package.

  For example:

  ```
  errCodeInt := C.libbpf_api_function()
  if errCodeInt != 0 {
    log.Errorf("uh oh: %s\n", syscall.Errno(errCodeInt))
  }
  ```

- APIs that return a pointer should be checked for
  NULL which indicates error. The error code is
  stored in errno. We can get the value of errno
  using the second return.

  For example:

  ```
  ptr, errno := C.libbpf_api_function()
  if ptr == nil {
    log.Errorf("uh oh: %s\n", errno)
  }
  ```

  We can also check if errno corresponds with a specific
  error (it implements the standard error interface).

  For example:

  ```
  ptr, errno := C.libbpf_api_function()
  if ptr == nil {
    if errno.Is(syscall.ENODENT) {
       // handle accordingly
    } else {
      log.Errorf("uh oh: %s\n", errno)
    }
  }

  ```

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -297,17 +297,6 @@ func bumpMemlockRlimit() error {
 	return nil
 }
 
-func errptrError(ptr unsafe.Pointer, format string, args ...interface{}) error {
-	negErrno := C.PTR_ERR(ptr)
-	errno := syscall.Errno(-int64(negErrno))
-	if errno == 0 {
-		return fmt.Errorf(format, args...)
-	}
-
-	args = append(args, errno.Error())
-	return fmt.Errorf(format+": %v", args...)
-}
-
 type NewModuleArgs struct {
 	KConfigFilePath string
 	BTFObjPath      string
@@ -348,9 +337,9 @@ func NewModuleFromFileArgs(args NewModuleArgs) (*Module, error) {
 		defer C.free(unsafe.Pointer(kConfigFile))
 	}
 
-	obj := C.bpf_object__open_file(bpfFile, &opts)
-	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
-		return nil, errptrError(unsafe.Pointer(obj), "failed to open BPF object %s", args.BPFObjPath)
+	obj, errno := C.bpf_object__open_file(bpfFile, &opts)
+	if obj == nil {
+		return nil, fmt.Errorf("failed to open BPF object %s: %s", args.BPFObjPath, errno)
 	}
 
 	return &Module{
@@ -390,9 +379,9 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 		defer C.free(unsafe.Pointer(kConfigFile))
 	}
 
-	obj := C.bpf_object__open_mem(bpfBuff, bpfBuffSize, &opts)
-	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
-		return nil, errptrError(unsafe.Pointer(obj), "failed to open BPF object %s: %v", args.BPFObjName, args.BPFObjBuff[:20])
+	obj, errno := C.bpf_object__open_mem(bpfBuff, bpfBuffSize, &opts)
+	if obj == nil {
+		return nil, fmt.Errorf("failed to open BPF object %s: %v: errno", args.BPFObjName, args.BPFObjBuff[:20], errno)
 	}
 
 	C.free(bpfBuff)
@@ -428,9 +417,9 @@ func (m *Module) Close() {
 }
 
 func (m *Module) BPFLoadObject() error {
-	ret := C.bpf_object__load(m.obj)
-	if ret != 0 {
-		return fmt.Errorf("failed to load BPF object")
+	cErr := C.bpf_object__load(m.obj)
+	if cErr != 0 {
+		return fmt.Errorf("failed to load BPF object", syscall.Errno(int(-cErr)))
 	}
 
 	return nil
@@ -438,10 +427,10 @@ func (m *Module) BPFLoadObject() error {
 
 func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	cs := C.CString(mapName)
-	bpfMap := C.bpf_object__find_map_by_name(m.obj, cs)
+	bpfMap, errno := C.bpf_object__find_map_by_name(m.obj, cs)
 	C.free(unsafe.Pointer(cs))
 	if bpfMap == nil {
-		return nil, fmt.Errorf("failed to find BPF map %s", mapName)
+		return nil, fmt.Errorf("failed to find BPF map %s: %s", mapName, errno)
 	}
 
 	return &BPFMap{
@@ -457,7 +446,7 @@ func (b *BPFMap) Pin(pinPath string) error {
 	errC := C.bpf_map__pin(b.bpfMap, path)
 	C.free(unsafe.Pointer(path))
 	if errC != 0 {
-		return fmt.Errorf("failed to pin map %s to path %s", b.name, pinPath)
+		return fmt.Errorf("failed to pin map %s to path %s: %s", b.name, pinPath, syscall.Errno(int(errC)))
 	}
 	return nil
 }
@@ -467,7 +456,7 @@ func (b *BPFMap) Unpin(pinPath string) error {
 	errC := C.bpf_map__unpin(b.bpfMap, path)
 	C.free(unsafe.Pointer(path))
 	if errC != 0 {
-		return fmt.Errorf("failed to unpin map %s from path %s", b.name, pinPath)
+		return fmt.Errorf("failed to unpin map %s from path %s: %s", b.name, pinPath, syscall.Errno(int(errC)))
 	}
 	return nil
 }
@@ -477,7 +466,7 @@ func (b *BPFMap) SetPinPath(pinPath string) error {
 	errC := C.bpf_map__set_pin_path(b.bpfMap, path)
 	C.free(unsafe.Pointer(path))
 	if errC != 0 {
-		return fmt.Errorf("failed to set pin for map %s to path %s", b.name, pinPath)
+		return fmt.Errorf("failed to set pin for map %s to path %s: %s", b.name, pinPath, syscall.Errno(int(errC)))
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -428,20 +417,20 @@ func (m *Module) Close() {
 }
 
 func (m *Module) BPFLoadObject() error {
-	ret := C.bpf_object__load(m.obj)
-	if ret != 0 {
-		return fmt.Errorf("failed to load BPF object")
+	cErr := C.bpf_object__load(m.obj)
+	if cErr != 0 {
+		return fmt.Errorf("failed to load BPF object", syscall.Errno(int(-cErr)))
```
rafaeldtinoco, line 422: Is this right ? Follow my thoughts:

You call bpf_object__load(NULL)... then you will face:

```C
if (!attr)
        return libbpf_err(-EINVAL);
```

and `libbpf_err(-EINVAL):`:

```C
static inline int libbpf_err(int ret)
{
        if (ret < 0)
                errno = -ret;
        return ret;
}
```

So it will return a negative number, but errno will still be positive on the error return. Why do you need to set it to negative again ? And also it is supposed to be an unsigned integer:

```go
type Errno uintptr
```

and all general errnos are positive as well:

```go
// Errors
const (
        E2BIG           = Errno(0x7)
        EACCES          = Errno(0xd)
        EADDRINUSE      = Errno(0x62)
        EADDRNOTAVAIL   = Errno(0x63)
        EADV            = Errno(0x44)
        EAFNOSUPPORT    = Errno(0x61)
        EAGAIN          = Errno(0xb)
        EALREADY        = Errno(0x72)
        EBADE           = Errno(0x34)
        EBADF           = Errno(0x9)
        EBADFD          = Errno(0x4d)
        EBADMSG         = Errno(0x4a)
        EBADR           = Errno(0x35)
        EBADRQC         = Errno(0x38)
        EBADSLT         = Errno(0x39)
        EBFONT          = Errno(0x3b)
        EBUSY           = Errno(0x10)
        ECANCELED       = Errno(0x7d)
        ECHILD          = Errno(0xa)
        ECHRNG          = Errno(0x2c)
        ECOMM           = Errno(0x46)
        ECONNABORTED    = Errno(0x67)
        ECONNREFUSED    = Errno(0x6f)
        ECONNRESET      = Errno(0x68)
        EDEADLK         = Errno(0x23)
...
```

rafaeldtinoco: Also, shouldn't you have a %s to stringify the errno ? Like shown at:

```go
func (e Errno) Error() string {
        if 0 <= int(e) && int(e) < len(errors) {
                s := errors[e]
                if s != "" {
                        return s
                }
        }
        return "errno " + itoa.Itoa(int(e))
}
``` 
Was that your intent here ?

rafaeldtinoco: There are other places doing the same thing. I'll clarify this first before moving on the review (as I'd like to understand the reasoning of that).

grantseltzer: Ok you are correct, I didn't mean to have negatives there, that was an inconsistency and mistake :-) 

Also yes, %s should be used since the `error` interface implements stringer. Just missing the %s here.

grantseltzer:
Relevant mailing list question: https://lore.kernel.org/bpf/CAO658oXGs3R=mQZo=aOEvRgG6O-obFbwR54V+QSn36uecF0+=w@mail.gmail.com/

derekparker:
Hey @grantseltzer I think there's some inconsistencies with the error handling in libbpf (at least that's what I'm observing).

If you check the other PR I submitted (https://github.com/aquasecurity/libbpfgo/pull/101) you can see `errC == -1` however `errno == ENOENT` (which is 2, not 1). libbpf seems to use ENOENT to indicate reading from an empty map, so it's valuable to have that information from the callers perspective to know it wasn't a "real" error, just an empty map with nothing to be read (at least that's how I understand it).

derekparker:
Also I believe some of Andrii's comments to be mistaken in his review on your documentation PR: https://patchwork.kernel.org/project/netdevbpf/patch/20211220054048.54845-1-grantseltzer@gmail.com/.

He claims that certain buffers for input buffers, such as the `keys` buffer argument to `lookup_and_delete_batch` or even just `delete_batch` but that's not the case. Your original documentation is correct in that the buffer is filled in by the call to the bpf batch operation, not used as an input set of keys (if that makes sense). This can be verified by looking at the kernel tests themselves: https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/testing/selftests/bpf/map_tests/htab_map_batch_ops.c

grantseltzer:
@derekparker Are you saying that `keys` is an output parameter for `bpf_map_delete_batch()`? I don't think that's the case. I do think it's the case for `bpf_map_lookup_and_delete_batch()` though.

derekparker:
> @derekparker Are you saying that `keys` is an output parameter for `bpf_map_delete_batch()`? I don't think that's the case. I do think it's the case for `bpf_map_lookup_and_delete_batch()` though.

Ack my mistake I think you're right.

derekparker:
@grantseltzer in my opinion, based on what I've seen during implementation (return value differing from errno in some cases) I think what this library should do is return the errno error wrapped in another error, so that callers can use patterns such as:

```
err := libbpfgo.SomeFunction()
if errors.Is(err, os.ErrNotExist) {
        // ...
}
```

That would mean in the implementation we would use a pattern of:

```
ret, errno := C.bpf_do_some_stuff()
if errno != syscall.Errno(0) {
        return fmt.Errorf("unable to do some bpf stuff: %w")
}
```

Note the `%w` to explicitly wrap the error instead of shadowing it.

Thoughts? I could test this change locally against my codebase but I'm pretty sure this would be sufficient and would standardize errors.

grantseltzer:
@derekparker I think that's a good idea, makes a lot of sense. I believe the inconsistencies between returned error and errno values is going to be solved by the time libbpf 1.0 is released, but i'm not entirely sure. Will verify and report back, getting back into the swing of things after vacation at the moment! I hope you had a nice holiday!

derekparker:
> @derekparker I think that's a good idea, makes a lot of sense. I believe the inconsistencies between returned error and errno values is going to be solved by the time libbpf 1.0 is released, but i'm not entirely sure. Will verify and report back, getting back into the swing of things after vacation at the moment! I hope you had a nice holiday!

Ack, thanks for the response! And likewise, still getting my head back in the game, haha. hope you had a nice holiday as well!

libbpfgo.go
```diff
@@ -672,7 +661,7 @@ func (b *BPFMap) UpdateBatch(keys, values unsafe.Pointer, count uint32) error {
 	}
 	errC := C.bpf_map_update_batch(b.fd, keys, values, &countC, bpfMapBatchOptsToC(&opts))
 	if errC != 0 {
-		return fmt.Errorf("failed to update map %s: %v", b.name, errC)
+		return fmt.Errorf("failed to update map %s: %w", b.name, syscall.Errno(int(errC)))
```
derekparker, line 664: I think we should wrap the errno value directly, due to inconsistencies with the returned value and errno in libbpf.

derekparker:
Only one comment otherwise this LGTM. I just think we should explicitly check errno and wrap that to not have to deal with inconsistencies between ret value and errno value.

derekparker (approved):
LGTM

rafaeldtinoco (approved):
LGTM

libbpfgo.go
```diff
@@ -1423,8 +1412,8 @@ func (hook *TcHook) Detach(tcOpts *TcOpts) error {
 func (hook *TcHook) Query(tcOpts *TcOpts) error {
 	opts := tcOptsToC(tcOpts)
 	ret := C.bpf_tc_query(hook.hook, opts)
-	if ret < 0 {
-		return syscall.Errno(-ret)
+	if ret != 0 {
+		return syscall.Errno(ret)
```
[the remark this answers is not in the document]
yanivagman: Errors are returned negative. See https://github.com/libbpf/libbpf/blob/e99f34e14447866bcbc38944387531c36f6c85fe/src/libbpf_internal.h#L454

And https://github.com/libbpf/libbpf/blob/21f90f61b0849ae654b7c78ba9ce34bfb74ce6f2/src/netlink.c#L756

libbpfgo.go
```diff
@@ -348,9 +338,9 @@ func NewModuleFromFileArgs(args NewModuleArgs) (*Module, error) {
 		defer C.free(unsafe.Pointer(kConfigFile))
 	}
 
-	obj := C.bpf_object__open_file(bpfFile, &opts)
-	if C.IS_ERR_OR_NULL(unsafe.Pointer(obj)) {
-		return nil, errptrError(unsafe.Pointer(obj), "failed to open BPF object %s", args.BPFObjPath)
+	obj, errno := C.bpf_object__open_file(bpfFile, &opts)
+	if obj == nil {
```
yanivagman, line 342: This will only be null if LIBBPF_STRICT_CLEAN_PTRS is set.
https://github.com/libbpf/libbpf/blob/e99f34e14447866bcbc38944387531c36f6c85fe/src/libbpf_internal.h#L476
Is that the case?

libbpfgo.go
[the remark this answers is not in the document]
rafaeldtinoco: The discussion about this was made at:

https://github.com/aquasecurity/libbpfgo/pull/100#discussion_r771589769

I wonder if I missed something there.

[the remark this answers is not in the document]
yanivagman: > You call bpf_object__load(NULL)... then you will face:
> 
> ```c
> if (!attr)
>         return libbpf_err(-EINVAL);
> ```
> 
> and `libbpf_err(-EINVAL):`:
> 
> ```c
> static inline int libbpf_err(int ret)
> {
>         if (ret < 0)
>                 errno = -ret;
>         return ret;
> }
> ```
> 
> So it will return a negative number, but errno will still be positive on the error return. Why do you need to set it to negative again ?

Like you write here, the return is a negative number, meaning ret is negative. To have a positive value, we should `syscall.Errno(-ret)`

PR #101, opened 2021-12-21
libbpfgo: Improve error message in batch operations

derekparker:
This patch adds support for improved error messages in batch operations.
It seemed as if the return code would be sufficient, but that seems not
to be the case. It turns out that we actually need the errno information
to further figure out what went wrong and communicate that to the
caller.

For example, for batch read (or batch read and delete) operations a
errno value of ENOENT can indicate that the map is empty, and that is
information that should be propogated back to the caller.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -642,9 +642,12 @@ func (b *BPFMap) GetValueAndDeleteBatch(keys, startKey, nextKey unsafe.Pointer, 
 		Flags:     C.BPF_ANY,
 	}
 
-	errC := C.bpf_map_lookup_and_delete_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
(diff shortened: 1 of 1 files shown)


derekparker:
@grantseltzer seems you were correct on the last PR when you suggested to check the errno value as well, my bad!

I'm submitting this initial patch as is in an effort to get feedback on if this is a good pattern to follow for the rest of the operations. Once that is settled I will update this patch to add errno checking for the other batch operations where necessary.

grantseltzer:
I have this included in my PR here: https://github.com/aquasecurity/libbpfgo/pull/100/files#diff-c9fafdb5381ed3544be49e84b94f3e3342e61382a849cb5b4bf674765ad27a20L647

but I don't have a specific check for ENOENT, do you think that should be handled different from different errors? If you agree this is a duplicate PR let's discuss in #100

Just to note, both the actual return from `bpf_map_lookup_and_delete_batch` and errno afterwards (in the case of an error) will contain the same value, so it isn't necessary to pull the value of errno with the double return value (see linked line above for example)

derekparker:
Going to move the discussion over there.

PR #103, opened 2022-01-06
helpers/osinfo: add security lockdown type detection

rafaeldtinoco:
This allows user to check '/sys/kernel/security/lockdown' status.

### DIFF
--- a/helpers/osinfo.go
+++ b/helpers/osinfo.go
@@ -242,3 +242,47 @@ func FtraceEnabled() (bool, error) {
 	}
 	return b[0] == '1', nil
 }
+
+type LockdownMode int32
+
+func (l LockdownMode) String() string {
+	return lockdownModeToString[l]
+}
+
+const (
+	NOVALUE LockdownMode = iota
+	NONE
+	INTEGRITY
+	CONFIDENTIALITY
+)
+
+var stringToLockdownMode = map[string]LockdownMode{
+	"none":            NONE,
+	"integrity":       INTEGRITY,
+	"confidentiality": CONFIDENTIALITY,
+}
+
+var lockdownModeToString = map[LockdownMode]string{
+	NONE:            "none",
+	INTEGRITY:       "integrity",
+	CONFIDENTIALITY: "confidentiality",
+}
+
+func Lockdown() (LockdownMode, error) {
+	LockdownFile := "/sys/kernel/security/lockdown"
+	data, err := os.ReadFile(LockdownFile)
+	if err != nil {
+		return NOVALUE, err
+	}
+
+	dataString := string(data[:])
+
+	for lockString, lockMode := range stringToLockdownMode {
+		tempString := fmt.Sprintf("[%s]", lockString)
+		if strings.Contains(dataString, tempString) {
+			return lockMode, nil
+		}
+	}
+
+	return NOVALUE, fmt.Errorf("could not get lockdown mode")
+}


rafaeldtinoco:
@grantseltzer this is the feature that will be added to tracee:

```go
// kernel lockdown check

lockdown, err := helpers.Lockdown()
if err != nil {
	return err
}
if lockdown == helpers.CONFIDENTIALITY {
	return fmt.Errorf("kernel lockdown is set to 'confidentiality', can't load eBPF programs.")
}
if debug {
	fmt.Fprintf(os.Stdout, "OSInfo: Security Lockdown is '%v'\n", lockdown)
}
```

After we're able to compile tracee with libbpfgo HEAD.

helpers/osinfo.go
```diff
@@ -242,3 +242,43 @@ func FtraceEnabled() (bool, error) {
...
+
+var lockdownModeToString = map[LockdownMode]string{
+	NONE:            "none",
+	INTEGRITY:       "integrity",
+	CONFIDENTIALITY: "confidentiality",
+}
+
+func Lockdown() (LockdownMode, error) {
+	LockdownFile := "/sys/kernel/security/lockdown"
+	data, _ := os.ReadFile(LockdownFile)
```
grantseltzer, line 273: I think we want to check this error

helpers/osinfo.go
```diff
@@ -242,3 +242,43 @@ func FtraceEnabled() (bool, error) {
...
+	dataString := string(data[:])
+
+	for lockString, lockMode := range stringToLockdownMode {
```
grantseltzer, line 276: Since maps aren't guaranteed to be in sequential order, is that an issue here? Can multiple values be present? (see my comment about what my file looks like)

grantseltzer (commented):
I'm not familiar with lockdown mode at all, but the content of my file looks like this:

```
[*] cat /sys/kernel/security/lockdown
[none] integrity confidentiality
```

Where do you find the schema for this file? Is it the same on every distribution?

helpers/osinfo.go
rafaeldtinoco: if data is anything else we will default to NOVALUE and error at the end, no ?

helpers/osinfo.go
rafaeldtinoco: the file syntax is always the 3 values in the same order and the one being used is with [].

rafaeldtinoco: Check https://wiki.archlinux.org/title/security#Kernel_lockdown_mode out and the lockdown manual page. There are some online discussions about it and some other projects, libe bpftrace, doing the same thing .

They go beyond and try to load small bpfs and create maps to test if other types of blockers are in place iirc.

grantseltzer: Ah got it, looks good then.

helpers/osinfo.go
grantseltzer: True, but I think just in case the code changes it would be better to return NOVALUE explicitly here.

grantseltzer (approved):

helpers/osinfo.go
simar7: Since we pass a NOVALUE and an error back, in the call site we also need to check both. Today we just check the error to be not nil and bail out https://github.com/aquasecurity/tracee/blob/a5bc030ec862039d483eb966bff09b16356dd744/cmd/tracee-ebpf/main.go#L87-L90

This breaks when the distro does not have a lockdown file in the location specified.  Alpine 3.15 is an example. Although out of the box it has no lockdown, this check prevents tracee-ebpf from running. At least that's my understanding when trying it out.

rafaeldtinoco: You're right AND I opened this (in Tracee's side) https://github.com/aquasecurity/tracee/issues/1396. Idea is that we have to bind mount the file through Makefile.tracee-make (so you can execute tracee in the development container) AND, likely, not to fail if the file doesn't exist (like you say so). I have a PR handling the Makefile case, and need to land a fix in libbpfgo as well. I'll try to do that tomorrow.

rafaeldtinoco: Being addressed at https://github.com/aquasecurity/tracee/pull/1402

Issue #105, opened 2022-01-10
invalid libbpfgo-static target


mtcherni95:
Prerequisites: no `bpftool` installed
Steps to reproduce : 

1. run `make clean`
2. run `make libbpfgo-static`: this will fail as we do not have `bpftool` installed, however it will an create an empty `./output/vmlinux.h` file:
```
mkdir -p ./output
INFO: generating ./output/vmlinux.h from /sys/kernel/btf/vmlinux
/bin/sh: 3: btf: not found
make: *** [Makefile:94: output/vmlinux.h] Error 127
```
3. if we run again `make libbpfgo-static` then compilation will (wrongly) succeed because `./output/vmlinux.h` exists (created in step 2):
```
mkdir -p ./output/libbpf
CC="gcc" CFLAGS="-g -O2 -Wall -fpie" LD_FLAGS="" \
   make -C /home/ubuntu/dev/libbpfgo/libbpf/src \
        BUILD_STATIC_ONLY=1 \
        OBJDIR=/home/ubuntu/dev/libbpfgo/output/libbpf \
        DESTDIR=/home/ubuntu/dev/libbpfgo/output \
        INCLUDEDIR= LIBDIR= UAPIDIR= install
make[1]: Entering directory '/home/ubuntu/dev/libbpfgo/libbpf/src'
  MKDIR    staticobjs
  CC       bpf.o
  CC       btf.o
  CC       libbpf.o
  CC       libbpf_errno.o
  CC       netlink.o
  CC       nlattr.o
  CC       str_error.o
  CC       libbpf_probes.o
  CC       bpf_prog_linfo.o
  CC       xsk.o
  CC       btf_dump.o
  CC       hashmap.o
  CC       ringbuf.o
  CC       strset.o
  CC       linker.o
  CC       gen_loader.o
  CC       relo_core.o
  AR       libbpf.a
  INSTALL  bpf.h libbpf.h btf.h libbpf_common.h libbpf_legacy.h xsk.h bpf_helpers.h bpf_helper_defs.h bpf_tracing.h bpf_endian.h bpf_core_read.h skel_internal.h
  INSTALL  libbpf.pc
  INSTALL  libbpf.a
make[1]: Leaving directory '/home/ubuntu/dev/libbpfgo/libbpf/src'
CC=clang \
        CGO_CFLAGS="-I/home/ubuntu/dev/libbpfgo/output" \
[...]

PR #107, opened 2022-01-10
checking if bpftool is installed

mtcherni95:
Solves: #105

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,10 @@ $(VMLINUXH): $(OUTPUT)
 		exit 1; \
 	fi
 	@if [ ! -f $(VMLINUXH) ]; then \
+		if [ ! $(BPFTOOL) ]; then \
+			echo "ERROR: could not find bpftool"; \
+			exit 1; \
+		fi; \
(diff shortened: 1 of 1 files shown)


rafaeldtinoco (commented):
I want to change libbpfgo Makefile to something like we just did to tracee's Makefile.one checks: https://github.com/aquasecurity/tracee/blob/main/Makefile.one#L39

WDYT ? We can then check tools, libs and compilers versions like we're doing there. I need to update libbpfgo Makefile with similar changes (and to create a single Makefile) in libbpfgo.

@grantseltzer what do you think ?

With that change, we can make a build target to depend on "tools and versions checking" per make target.

grantseltzer:
@rafaeldtinoco Your makefile wizardy would be great to have in libbpfgo, but for the sake of getting this change in before the next release (i'd like to cut it today) I think @mtcherni95's fix works well for now. Is that ok?

rafaeldtinoco:
@grantseltzer its perfect and I agree. Its better to have this fix now then. I can work the Makefile for the next release (or even after Makefile.one is ready in tracee). Cheers!

rafaeldtinoco (approved):
LGTM

PR #109, opened 2022-01-11
Fix missing errno checks

grantseltzer:


yanivagman (approved):
LGTM.
Just please verify the errors are returned as expected (only need to check one case for each):
1. Where IS_ERR_OR_NULL was previously used
2. For negative return values - check that the correct syscall.Errno is returned

PR #111, opened 2022-01-11
Only check errno when error condition hit

derekparker:
It's possible to get an errno value from a successful libbpf call as
errno isn't reset before each CGO call. Instead of checking errno
unconditionally after each libbpf call, only check it once we've
determined we have actually hit an error condition.

This fixes a bunch of breakages and can be verified by running the
selftests and ensuring they all pass.


derekparker:
Closing in favor of https://github.com/aquasecurity/libbpfgo/pull/112 which has this and another commit to fix another issue.

PR #112, opened 2022-01-13
Handle errno better and enoent better for batch operations

derekparker:
Details in the commits but essentially we should only check errno if we detect an error condition (commit 1) and we should also still return information from batch operations even if we get `ENOENT` because that just means we've read everything. Not returning the information to the user will cause that data to be lost.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -608,14 +608,19 @@ func (b *BPFMap) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Poi
 		Flags:     C.BPF_ANY,
 	}
 
-	_, errno := C.bpf_map_lookup_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
-	if errno != nil {
-		return nil, fmt.Errorf("failed to batch lookup values %v in map %s: %w", keys, b.name, errno)
+	var reterr error
+	ret, errno := C.bpf_map_lookup_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
+	if ret < 0 && errno != nil {
+		if errors.Is(errno, syscall.ENOENT) {
+			reterr = errno
+		} else {
+			return nil, fmt.Errorf("failed to batch lookup values %v in map %s: %w", keys, b.name, errno)
+		}
 	}
 
 	parsedVals := collectBatchValues(values, count, b.ValueSize())
 
-	return parsedVals, nil
+	return parsedVals, reterr
 }
 
 // GetValueAndDeleteBatch allows for batch lookups of multiple keys and deletes those keys.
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -339,7 +339,7 @@ func NewModuleFromFileArgs(args NewModuleArgs) (*Module, error) {
 	}
 
 	obj, errno := C.bpf_object__open_file(bpfFile, &opts)
-	if errno != nil {
+	if obj == nil && errno != nil {
```
grantseltzer, line 342: Why is checking errno not already enough to make sure there's an error? It seems that Go does clear errno before each C function call.

yanivagman: Not related to the comment of @grantseltzer 
libbpf might return a non-null value in some cases.
For example: https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L7080.
In that case, we will mistakenly return that no error occured

libbpfgo.go
```diff
@@ -381,7 +381,7 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 	}
 
 	obj, errno := C.bpf_object__open_mem(bpfBuff, bpfBuffSize, &opts)
-	if errno != nil {
+	if obj == nil && errno != nil {
```
yanivagman, line 384: same

libbpfgo.go
derekparker: Even if Go does clear it before the call it's still possible that something within the call sets `errno` but we still don't actually have an error condition, the call that we originally made did succeed.

Without this change a few of the selftests error out. Also in this patch is a fix for map batch operations where `errno` could equal `ENOENT`. This happens when we've read an entire map, however we may still have gotten some data back so this patch fixes things to ensure whatever data we got is still returned, even if it's less than the batch amount.

libbpfgo.go
```diff
@@ -430,7 +430,7 @@ func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	cs := C.CString(mapName)
 	bpfMap, errno := C.bpf_object__find_map_by_name(m.obj, cs)
 	C.free(unsafe.Pointer(cs))
-	if errno != nil {
+	if bpfMap == nil && errno != nil {
```
yanivagman, line 433: Here it is ok to check for null bpfMap:
https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L9394

This inconsistency in libbpf is really confusing. I think it should be fixed before 1.0

libbpfgo.go
```diff
@@ -1057,7 +1067,7 @@ func doAttachKprobeLegacy(prog *BPFProg, kp string, isKretprobe bool) (*BPFLink,
 	cbool := C.bool(isKretprobe)
 	link, errno := C.attach_kprobe_legacy(prog.prog, cs, cbool)
 	C.free(unsafe.Pointer(cs))
-	if errno != nil {
+	if link == nil && errno != nil {
```
yanivagman, line 1070: This function is implemented in libbpfgo - it is ok to check for nil

libbpfgo.go
```diff
@@ -1094,7 +1104,7 @@ func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffe
 	}
 
 	rb, errno := C.init_ring_buf(bpfMap.fd, C.uintptr_t(slot))
-	if errno != nil {
+	if rb == nil && errno != nil {
```
yanivagman, line 1107: This function is implemented in libbpfgo - it is ok to check for nil

grantseltzer:
@derekparker Do you have thoughts on Yaniv's latest comments? Overall I like your approach!

derekparker:
> @derekparker Do you have thoughts on Yaniv's latest comments? Overall I like your approach!

Yeah, I'll update the code to take the feedback into account! Sorry for the delay.

grantseltzer:
No worries!

grantseltzer:
Hi @derekparker, we're running into stability issues in tracee with the current state of libbpfgo, caused by my original error handling changes. Even with your changes we still have failing tests. In #119 we decided to revert back to the original state of error handling so that we can return to stability while we figure error handling out properly, while simultaneously improving tests to avoid this issue in the future. 

The way forward would be to cherry pick the original commits that were reverted in #112 and base your changes on top of that, in addition to whatever fixes are needed on top of that. I understand that's a good amount of work and i'm not sure what your team's priorities are, so please feel free to let me handle this work if you'd like me to. Happy to hear your thoughts, as always!

derekparker:
@grantseltzer you can feel free to pick up this work if you'd like!

rafaeldtinoco:
I'm closing this for now as we're currently working in some other PR dealing with error handling.

PR #119, opened 2022-01-21
Main fixes

rafaeldtinoco:
@grantseltzer @itaysk @yanivagman 

My proposal here is that we do a step back for the error handling, so we can have libbpfgo main in a good shape. 

There is on going discussion at PR: #112 about the error handling approach (in another PR) and my feeling is that we're not ready for that change yet (main return behavior for important functions is currently broken).

Would you mind going back to keep the branch stable and we're able to land other fixes (like the ones being proposed here) ?


grantseltzer (approved):
This LGTM for now, #112 will have to cherry pick the reverted commits to rebase on top of and we can test on top of that. Unfortunate that we have to do this, but I agree that it's the right call.

Issue #121, opened 2022-01-23
panic when using perf buffer with libbpf v0.6.1


yanivagman:
This behavior was observed when using Tracee with the latest changes (which bumped libbpfgo version and libbpf to v0.6.1).
Further details in https://github.com/aquasecurity/tracee/issues/1383

Issue #34, opened 2021-07-08
Remove kprobe legacy logic once upstream accepts it


rafaeldtinoco:
Discussing **libbpf: add legacy kprobe attach support** at:

https://github.com/libbpf/libbpf/issues/317

libbpfgo "cgo: attach_kprobe_legacy" and "doAttachKprobeLegacy" logic should be removed once libbpf accepts that patch.

rafaeldtinoco:
Latest proposal at [LORE](https://lore.kernel.org/bpf/20210625044459.1249282-1-rafaeldtinoco@gmail.com/) thread.

rafaeldtinoco:
Latest proposal at:

https://lore.kernel.org/bpf/20210801051123.3822498-1-rafaeldtinoco@gmail.com/

already to be applied on top of the:

[PATCH v3 bpf-next 00/14] BPF perf link and user-provided bpf_cookie

patch set from Andrii. It will need to wait until that patch set is merged. It correctly replaces libbpfgo legacy kprobe implementation embedded at libbpfgo.go source code (for current libbpf upstream branch).

rafaeldtinoco:
Commit got merged in for-next (bpf-next):

```
commit 467b3225553a (HEAD -> rafaeldtinoco, origin/master, origin/for-next, origin/HEAD)
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Sun Sep 12 03:48:44 2021

    libbpf: Introduce legacy kprobe events support

    Allow kprobe tracepoint events creation through legacy interface, as the
    kprobe dynamic PMUs support, used by default, was only created in v4.17.

    Store legacy kprobe name in struct bpf_perf_link, instead of creating
    a new "subclass" off of bpf_perf_link. This is ok as it's just two new
    fields, which are also going to be reused for legacy uprobe support in
    follow up patches.

    Signed-off-by: Rafael David Tinoco <rafaeldtinoco@gmail.com>
    Signed-off-by: Andrii Nakryiko <andrii@kernel.org>
    Link: https://lore.kernel.org/bpf/20210912064844.3181742-1-rafaeldtinoco@gmail.com
```

It will get into bpf-next and, eventually, into libbpf (github). We will be able to get rid of legacy kprobe functions.

rafaeldtinoco:
Alright, the legacy kprobe logic just got merged into libbpf:

[...]

PR #123, opened 2022-01-25
0.6.1 fixes

rafaeldtinoco:
commit d259d81
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Tue Jan 25 16:36:22 2022

    libbpfgo: use perf_buffer__new new API and fix style

    Fix: #121

    This commit forces libbpfgo.go C embedded source code function
    init_perf_buf to use the new perf_buffer__new() API. This is needed to
    make sure libbpf doesn't choose to execute compatible perf_buffer__new()
    function prototype, causing execution issues (seg faults) in some
    environments.

commit 1a666df
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Tue Jan 25 12:14:30 2022

    kprobe: remove legacy kprobes and unneeded includes

    Fix: #34

    Since libbpf commit 749b394 ("libbpf: Introduce legacy kprobe events
    support"), and subsequent fixes and refactorings to this code, there is
    no longer a need for libbpfgo to keep its own legacy kprobes code, as
    libbpfgo is now at libbpf v0.6.1 and that logic was added since v0.6.0.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -11,33 +11,31 @@ package libbpfgo
 #include <bpf/libbpf.h>
 
 #ifndef MAX_ERRNO
-#define MAX_ERRNO       4095
+#define MAX_ERRNO           4095
+#define IS_ERR_VALUE(x)     ((x) >= (unsigned long)-MAX_ERRNO)
 
-#define IS_ERR_VALUE(x) ((x) >= (unsigned long)-MAX_ERRNO)
-
-static inline bool IS_ERR(const void *ptr) {
-	return IS_ERR_VALUE((unsigned long)ptr);
+static inline bool IS_ERR(const void *ptr)
+{
+    return IS_ERR_VALUE((unsigned long)ptr);
 }
 
-static inline bool IS_ERR_OR_NULL(const void *ptr) {
-	return !ptr || IS_ERR_VALUE((unsigned long)ptr);
+static inline bool IS_ERR_OR_NULL(const void *ptr)
+{
+    return !ptr || IS_ERR_VALUE((unsigned long)ptr);
 }
 
-static inline long PTR_ERR(const void *ptr) {
-	return (long) ptr;
+static inline long PTR_ERR(const void *ptr)
+{
+    return (long) ptr;
 }
 #endif
 
-extern void perfCallback(void *ctx, int cpu, void *data, __u32 size);
-extern void perfLostCallback(void *ctx, int cpu, __u64 cnt);
-
-extern int ringbufferCallback(void *ctx, void *data, size_t size);
-
-int libbpf_print_fn(enum libbpf_print_level level,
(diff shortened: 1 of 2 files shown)


libbpfgo.go
```diff
@@ -3,77 +3,80 @@ package libbpfgo
...
 
-#include <asm-generic/unistd.h>
-#include <errno.h>
-#include <fcntl.h>
-#include <linux/perf_event.h>
-#include <linux/unistd.h>
-#include <string.h>
-#include <unistd.h>
+#include <bpf/bpf.h>
+#include <bpf/libbpf.h>
```
yanivagman, line 12: Are these changes related to this commit or the other one(removing the kprobe)?
We better put it in the relevant commit for the sake of git history

libbpfgo.go
```diff
@@ -3,77 +3,80 @@ package libbpfgo
...
 
 #ifndef MAX_ERRNO
 #define MAX_ERRNO       4095
 
 #define IS_ERR_VALUE(x) ((x) >= (unsigned long)-MAX_ERRNO)
 
-static inline bool IS_ERR(const void *ptr) {
-	return IS_ERR_VALUE((unsigned long)ptr);
+static inline bool IS_ERR(const void *ptr)
+{
```
yanivagman, line 20: Why changing to this convention?
In our code we keep the opening brace in the function signature line

libbpfgo.go
yanivagman: No need to create a new one. I just think it is more relevant for the other commit that removed the kprobes code, isn't it?

libbpfgo.go
rafaeldtinoco: Just so you don't think I didn't put some thoughts on this before suggesting:

This is K&R style ([kernel variant](https://en.wikipedia.org/wiki/Indentation_style#Variant:_Linux_kernel)), which I think we should start complying to in our C sources.

Reason: With K&R style we can check formatting of this, and our eBPF source codes, with tools such as check_patch, during the PRs (Not particularly on this case, but, this code should also be extracted from .go code if possible (near future) so we can check formatting in PRs).

In this case, since it is part of golang src file, it has tabs expanded to 4 chars, which shouldn't be true for a C src file, but doesn't make sense to have different tab expansion policy in a go file.

If you don't like it, we can move back to previous style (with fixes to mixed tab and space indendation it had here and there). I just consider an error treating C code with a go formatting pattern (as well as having +100 cols lines in golang src, which is something we should try to avoid to improve code readability).

WDYT ?

But I'd like you to request for you to consider K&R kernel style not for this, but also for the eBPF code refactoring within tracee.

yanivagman: Ok, that's a good cause.
I'm ok with this change

yanivagman (approved):

PR #126, opened 2022-01-31
Compile tracee submodule against local libbpfgo

grantseltzer:
This adds a make target to compile a tracee submodule against local libbpfgo. Since tracee is libbpfgo's most notable consumer, it's a good smoke test to make sure local changes don't break tracee. Once e2e tests are in a good spot we can run those from here as well.

### DIFF
--- a/testing/compile-tracee.sh
+++ b/testing/compile-tracee.sh
@@ -1,0 +1,53 @@
+#!/bin/bash -e
+
+#
+# This script downloads and updates the aquasecurity/tracee repository
+# 
+
+die() {
+    echo ${@}
+    exit 1
+}
+
+BASEDIR=$(dirname "${0}") ; cd ${BASEDIR}/../ ; BASEDIR=$(pwd) ; cd ${BASEDIR}
+TRACEE_REPO="https://github.com/aquasecurity/tracee.git"
+TRACEE_DIR="${BASEDIR}/testing/tracee"
+RETURN_CODE=0
+
+CMDS="rsync git cp rm mv"
+for cmd in ${CMDS}; do
+    command -v $cmd 2>&1 >/dev/null || die "cmd ${cmd} not found"
+done
+
+branch_clean() {
+    cd ${1} || die "could not change dirs"
+
+    # small sanity check
+    [ ! -f ./Readme.md ] && die "$(basename $(pwd)) not a repo dir"
+
+    git fetch -a || die "could not fetch ${1}"  # make sure its updated
+    git clean -fdX                              # clean leftovers
+    git reset --hard                            # reset letfovers
+    git checkout origin/main -b main-$$
+    git branch -D main
+    git branch -m main-$$ main                  # origin/main == main
+
+    cd ${BASEDIR}
+}
+
+[ ! -d ${TRACEE_DIR} ] && git clone "${TRACEE_REPO}" ${TRACEE_DIR}
+if [ -z ${SKIP_FETCH} ]; then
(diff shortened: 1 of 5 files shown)


rafaeldtinoco:
I'm not sure I like this approach @grantseltzer. I don't think we should include tracee as a submodule of libbpfgo.

I mean, we can automate a test (shell ?) to clone the repo, test things out in an git-ignored directory... like a selftest... but having it as a submodule doesn't seem like a good idea to me.

Check out `3rdparty/btfhub.sh` file in tracee for an idea how to clone and keep an ignored directory updated (for btfhub in that case). Maybe we could do the same for this test. 

I just don't want to disturb other projects who have nothing to do with tracee. WDYT ?

rafaeldtinoco (changes requested):
I'm not sure I like this approach @grantseltzer. I don't think we should include tracee as a submodule of libbpfgo.

I mean, we can automate a test (shell ?) to clone the repo, test things out in an git-ignored directory... like a selftest... but having it as a submodule doesn't seem like a good idea to me.

Check out 3rdparty/btfhub.sh file in tracee for an idea how to clone and keep an ignored directory updated (for btfhub in that case). Maybe we could do the same for this test.

I just don't want to disturb other projects who have nothing to do with tracee. WDYT ?

grantseltzer:
> I'm not sure I like this approach @grantseltzer. I don't think we should include tracee as a submodule of libbpfgo.
> 
> I mean, we can automate a test (shell ?) to clone the repo, test things out in an git-ignored directory... like a selftest... but having it as a submodule doesn't seem like a good idea to me.
> 
> Check out `3rdparty/btfhub.sh` file in tracee for an idea how to clone and keep an ignored directory updated (for btfhub in that case). Maybe we could do the same for this test.
> 
> I just don't want to disturb other projects who have nothing to do with tracee. WDYT ?

That makes sense, submodule is tying it closer together, i'll redo it as just cloning in a script

testing/compile-tracee.sh
```diff
@@ -0,0 +1,53 @@
...
+TRACEE_DIR="${BASEDIR}/testing/tracee"
+RETURN_CODE=0
+
+CMDS="rsync git cp rm mv"
+for cmd in ${CMDS}; do
+    command -v $cmd 2>&1 >/dev/null || die "cmd ${cmd} not found"
+done
+
+branch_clean() {
+    cd ${1} || die "could not change dirs"
```
rafaeldtinoco, line 23:
```
++ basename /home/rafaeldtinoco/work/ebpf/libbpfgo/testing/tracee
+ die 'tracee not a repo dir'
+ echo tracee not a repo dir
tracee not a repo dir
```

This is not right, you are CWDing to ./testing/tracee and it does not contain Readme.md file.

testing/compile-tracee.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
+
+#
+# This script downloads and updates the aquasecurity/tracee repository
+#
```
rafaeldtinoco, line 5: nit: there is an extra space after # here.

rafaeldtinoco (changes requested):

testing/compile-tracee.sh
grantseltzer: I don't understand what you mean, doesn't it?

rafaeldtinoco: Nm, my local tree had garbage, sorry.

rafaeldtinoco (approved):
LGTM

PR #127, opened 2022-02-03
Error handling fixes

grantseltzer:
This takes a simpler approach to error handling compare to previous PRs that were reverted. The conclusions I made are mainly that the macro `IS_ERR_OR_NULL` works well for backwards/forward compatibility. The APIs that return pointers will return NULL on error with error code in errno in the 1.0 libbpf release. Since this isn't yet consistent, it's better to just check if the return is NULL or if it contains the error code. 

The main changes here are getting more information out of directly returned error codes (non pointer APIs) using `syscall.Errno()`.

I also added a selftest which checks that error codes are properly handled in a few different API functions. If reviewers like this approach, i'll add more checks in that selftest before merging.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -327,10 +327,10 @@ func (m *Module) BPFLoadObject() error {
 
 func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	cs := C.CString(mapName)
-	bpfMap := C.bpf_object__find_map_by_name(m.obj, cs)
+	bpfMap, errno := C.bpf_object__find_map_by_name(m.obj, cs)
 	C.free(unsafe.Pointer(cs))
 	if bpfMap == nil {
-		return nil, fmt.Errorf("failed to find BPF map %s", mapName)
+		return nil, fmt.Errorf("failed to find BPF map %s: %w", mapName, errno)
 	}
 
 	return &BPFMap{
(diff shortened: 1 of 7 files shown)


libbpfgo.go
```diff
@@ -327,10 +327,10 @@ func (m *Module) BPFLoadObject() error {
 
 func (m *Module) GetMap(mapName string) (*BPFMap, error) {
 	cs := C.CString(mapName)
-	bpfMap := C.bpf_object__find_map_by_name(m.obj, cs)
+	bpfMap, errno := C.bpf_object__find_map_by_name(m.obj, cs)
```
grantseltzer, line 330: This API function is already returning just NULL on error with error code in errno.

libbpfgo.go
```diff
@@ -677,10 +677,10 @@ func (it *BPFMapIterator) Err() error {
 
 func (m *Module) GetProgram(progName string) (*BPFProg, error) {
 	cs := C.CString(progName)
-	prog := C.bpf_object__find_program_by_name(m.obj, cs)
+	prog, errno := C.bpf_object__find_program_by_name(m.obj, cs)
```
grantseltzer, line 680: This API function is already returning just NULL on error with error code in errno.

libbpfgo.go
rafaeldtinoco: Cool, I came exactly because of that. Was going to ask why those 2 calls were different than all others, having errno being set directly in cgo's call. Thanks.

PR #128, opened 2022-02-13
helpers: Added dynamic symbol resolver

guyarb:
Dynamic symbols are very important to attach u(ret)probe on common libraries such as openssl, libc, etc.

issue: https://github.com/aquasecurity/libbpfgo/issues/129

### DIFF
--- a/helpers/elf.go
+++ b/helpers/elf.go
@@ -6,19 +6,9 @@ import (
 	"fmt"
 )
 
-// SymbolToOffset attempts to resolve a 'symbol' name in the binary found at
-// 'path' to an offset. The offset can be used for attaching a u(ret)probe
-func SymbolToOffset(path, symbol string) (uint32, error) {
-	f, err := elf.Open(path)
-	if err != nil {
-		return 0, fmt.Errorf("could not open elf file to resolve symbol offset: %w", err)
-	}
-
-	syms, err := f.Symbols()
-	if err != nil {
-		return 0, fmt.Errorf("could not open symbol section to resolve symbol offset: %w", err)
-	}
-
+// symbolToOffset attempts to resolve a 'symbol' name in the given binary to an offset.
+// The offset can be used for attaching a u(ret)probe.
+func symbolToOffset(symbol string, f *elf.File, syms []elf.Symbol) (uint32, error) {
 	sectionsToSearchForSymbol := []*elf.Section{}
 
 	for i := range f.Sections {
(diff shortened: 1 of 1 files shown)


helpers/elf.go
```diff
@@ -48,5 +38,37 @@ func SymbolToOffset(path, symbol string) (uint32, error) {
...
+	if err != nil {
+		return 0, fmt.Errorf("could not open symbol section to resolve symbol offset: %w", err)
+	}
+
+	return symbolToOffset(symbol, f, syms)
+}
+
+// DynamicSymbolToOffset attempts to resolve a dynamic 'symbol' name in the binary found at
+// 'path' to an offset. The offset can be used for attaching a u(ret)probe
+func DynamicSymbolToOffset(path, symbol string) (uint32, error) {
```
grantseltzer, line 62: What do you think about adding this logic to `SymbolToOffset` instead? If the named symbol isn't in the symbol table, check dynamic symbols. Is it possible to have a name collision?

guyarb: I'm in favor of your suggestion, I didn't want to "break" anyone using the original function, so ill change the code accordingly.

regarding the name collision - so the short answer would be yes, but it is only because the dynamic symbol table can contain exported functions that will appear in the regular symbol table.

a nice answer in stack overflow https://reverseengineering.stackexchange.com/a/21623

guyarb: created a simple SO (followed https://www.cprogramming.com/tutorial/shared-libraries-linux-gcc.html)
and checked both the regular symbol table and the dynamic symbol table
you can see the duplications
image

guyarb:
> This is great, thank you! Just one comment, let me know what you think!

Thanks!
commented + fixed

helpers/elf.go
```diff
@@ -48,5 +55,5 @@ func SymbolToOffset(path, symbol string) (uint32, error) {
 		}
 	}
 
-	return 0, fmt.Errorf("symbol not found")
+	return 0, errors.New("symbol not found")
```
grantseltzer, line 58:
```suggestion
	return 0, fmt.Errorf("symbol %s not found in %s", symbol, path)
```

helpers/elf.go
```diff
@@ -14,11 +14,18 @@ func SymbolToOffset(path, symbol string) (uint32, error) {
...
 
-	syms, err := f.Symbols()
-	if err != nil {
-		return 0, fmt.Errorf("could not open symbol section to resolve symbol offset: %w", err)
+	regularSymbols, regularSymbolsErr := f.Symbols()
+	dynamicSymbols, dynamicSymbolsErr := f.DynamicSymbols()
+
+	// Only if we failed getting both regular and dynamic symbols - then we abort.
+	if regularSymbolsErr != nil && dynamicSymbolsErr != nil {
+		return 0, fmt.Errorf("could not open symbol sections to resolve symbol offset: %w", regularSymbolsErr)
```
grantseltzer, line 22:
```suggestion
		return 0, fmt.Errorf("could not open symbol sections to resolve symbol offset: %w, %w", regularSymbolsErr, dynamicSymbolsError)
```

grantseltzer (changes requested):
Couple of small suggestions (let me know WYT) but otherwise LGTM!

helpers/elf.go
```diff
@@ -14,11 +14,18 @@ func SymbolToOffset(path, symbol string) (uint32, error) {
...
 
-	syms, err := f.Symbols()
-	if err != nil {
-		return 0, fmt.Errorf("could not open symbol section to resolve symbol offset: %w", err)
+	regularSymbols, regularSymbolsErr := f.Symbols()
+	dynamicSymbols, dynamicSymbolsErr := f.DynamicSymbols()
+
+	// Only if we failed getting both regular and dynamic symbols - then we abort.
+	if regularSymbolsErr != nil && dynamicSymbolsErr != nil {
+		return 0, fmt.Errorf("could not open symbol sections to resolve symbol offset: %w, %w", regularSymbolsErr, dynamicSymbolsError)
```
grantseltzer, line 22:
```suggestion
		return 0, fmt.Errorf("could not open symbol sections to resolve symbol offset: %w, %w", regularSymbolsErr, dynamicSymbolsErr)
```

grantseltzer (approved):
I messed up that error name woops lol, thanks for the contribution!

PR #133, opened 2022-02-22
introduce kernelsyms helper

itamarmaouda101:
Hi 

This PR will add `kernelsyms` helper that will give the ability to create a map of the known kernel symbols in the current Os.

Also this PR will help the [syscall detection PR in tracee](https://github.com/aquasecurity/tracee/pull/1394).

Btw the first idea were to implement it in Tracee but me and @yanivagman thought it will be better to include it

### DIFF
--- a/helpers/kernel_symbols.go
+++ b/helpers/kernel_symbols.go
@@ -1,0 +1,105 @@
+package helpers
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+/*
+ * The helpers in this file gives the ability to hold all the known kernel symbols.
+ * the package parse the /proc/kallsyms file that hold the known kernel symbol
+ *
+ * The KernelSymbolTable type holds map of all the kernel symbols with a key which is the kernel object owner and the name with under-case between them
+ * which means that symbolMap looks like [objectOwner_objectname{SymbolData}, objectOwner_objectname{SymbolData}, etc...]
+ * the key naming is because sometimes kernel symbols can have the same name or the same address which prevents to key the map with only one of them
+ *
+ */
+
+type KernelSymbolTable struct {
+	symbolMap   map[string]KernelSymbol
+	initialized bool
+}
+
+type KernelSymbol struct {
+	Name    string
+	Type    string
+	Address uint64
+	Owner   string
+}
+
+/* NewKernelSymbolsMap initiates  the kernel symbol map by parsing the /proc/kallsyms file.
+ * each line continas the symbol's address, segment type, name, module owner (which can be empty in case the symbol is owned by the system)
+ * Note: the key of the map is the symbol owner and the symbol name (with undercase between them)
+ */
+func NewKernelSymbolsMap() (*KernelSymbolTable, error) {
+	var KernelSymbols = KernelSymbolTable{}
+	KernelSymbols.symbolMap = make(map[string]KernelSymbol)
+	file, err := os.Open("/proc/kallsyms")
+	if err != nil {
+		return nil, fmt.Errorf("Could not open /proc/kallsyms")
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		line := strings.Fields(scanner.Text())
+		//if the line is less than 3 words, we can't parse it (one or more fields missing)
+		if len(line) < 3 {
+			continue
+		}
+		symbolAddr, err := strconv.ParseUint(line[0], 16, 64)
+		if err != nil {
+			continue
+		}
+		symbolName := line[2]
+		symbolType := line[1]
+		symbolOwner := "system"
+		//if the line is only 3 words then the symbol is owned by the system
+		if len(line) > 3 {
+			symbolOwner = line[3]
+		}
+		symbolKey := fmt.Sprintf("%s_%s", symbolOwner, symbolName)
+		KernelSymbols.symbolMap[symbolKey] = KernelSymbol{symbolName, symbolType, symbolAddr, symbolOwner}
+	}
+	KernelSymbols.initialized = true
+	return &KernelSymbols, nil
+}
+
+// TextSegmentContains checks if a given address is in the kernel text segment by compare it to the kernel text segment address boundaries
+func (k *KernelSymbolTable) TextSegmentContains(addr uint64) (bool, error) {
+	if !k.initialized {
+		return false, fmt.Errorf("KernelSymbolTable symbols map isnt initialized\n")
+	}
+	stext, err := k.GetSymbolByName("system", "_stext")
+	if err != nil {
+		return false, err
+	}
+	etext, err := k.GetSymbolByName("system", "_etext")
+	if err != nil {
+		return false, err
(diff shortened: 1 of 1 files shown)


helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+/*
+ * This package gives the ability to hold all the known kernel symbols.
```
grantseltzer, line 12: This is in the helpers package (which is good, don't need another one), but it would be more accurate to say "The helpers in this file gives the ability..."

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+		return false, err
+	}
+	etext, err := k.GetSymbolByName("system", "_etext")
+	if err != nil {
+		return false, err
+	}
+	if (addr >= stext.Address) && (addr < etext.Address) {
+		return true, nil
+	}
+	return false, nil
```
grantseltzer, line 48:
```suggestion
	return ((addr >= stext.Address) && (addr < etext.Address)), nil
```

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type KernelSymbol struct {
+	Name    string
+	Type    string
+	Address uint64
+	Owner   string
+}
+
+func (k *KernelSymbolTable) IsInTextSegment(addr uint64) (bool, error) {
```
grantseltzer, line 33: Since the symbol table itself is the method receiver I think it makes more sense for the method to be named `ContainsAddr`

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return false, err
+	}
+	if (addr >= stext.Address) && (addr < etext.Address) {
+		return true, nil
+	}
+	return false, nil
+}
+
+//GetSymbolByAddr returns a symbol by a given address
```
grantseltzer, line 51:
```suggestion
// GetSymbolByAddr returns a symbol by a given address
```

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+func (k *KernelSymbolTable) GetSymbolByAddr(addr uint64) (*KernelSymbol, error) {
+	for _, Symbol := range k.symbolMap {
+		if Symbol.Address == addr {
+			return &Symbol, nil
+		}
+	}
+	return nil, fmt.Errorf("symbol not found")
+}
+
+//GetSymbolByName returns a symbol by a given name and owner
```
grantseltzer, line 61:
```suggestion
// GetSymbolByName returns a symbol by a given name and owner
```

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//GetSymbolByAddr returns a symbol by a given address
+func (k *KernelSymbolTable) GetSymbolByAddr(addr uint64) (*KernelSymbol, error) {
+	for _, Symbol := range k.symbolMap {
+		if Symbol.Address == addr {
+			return &Symbol, nil
+		}
+	}
+	return nil, fmt.Errorf("symbol not found")
```
grantseltzer, line 58:
```suggestion
	return nil, fmt.Errorf("symbol not found at address: 0x%x", addr)
```

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//GetSymbolByName returns a symbol by a given name and owner
+func (k *KernelSymbolTable) GetSymbolByName(owner string, name string) (*KernelSymbol, error) {
+	key := fmt.Sprintf("%s_%s", owner, name)
+	symbol, exist := k.symbolMap[key]
+	if exist {
+		return &symbol, nil
+	}
+	return nil, fmt.Errorf("symbol not found")
```
grantseltzer, line 68:
```suggestion
	return nil, fmt.Errorf("symbol not found by name and owner: %s", key)
```

helpers/kernel_symbols.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// NewKernelSymbolsMap Initiate the kernel symbol map
```
grantseltzer, line 71:
```suggestion
// NewKernelSymbolsMap initiates the kernel symbol map
```

grantseltzer (changes requested):
A bunch of small nits, but otherwise this is great, nice work

helpers/kernel_symbols.go
itamarmaouda101: I think it better to create another function that will be member of `KernelSymbol struct` that will can compare the address, like that one:
`func (k *KernelSymbol) ContainsAddr(boundreyLowerAddr uint64, boundreyUpperAddr uint64) (bool, error) `

and ill change `IsInTextSegment` to use the ContainsAddr, like that.

what you think? @grantseltzer

rafaeldtinoco (approved):
LGTM, but, please, before merging:
1. make sure tracee change works with this version
2. squash all commits into a single one (since its a single change)
3. bump tracee's libbpfgo version dependency to this commit ID in the first commit of that PR of yours.
Thank you!

helpers/kernel_symbols.go
grantseltzer: I don't agree. I feel that the method name should be grammatically based off the method receiver. For example:

`KernelSymbolTable.IsInTextSegment()` doesn't make sense to me because you're not checking if the kernel symbol table is in the text segment, you're checking if the passed address is in the text segment. 

By the same logic, having `KernelSymbol.ContainsAddr()` doesn't make sense because you're not checking if the kernel symbol contains the address, you're checking if the kernel symbol is within the range of passed addresses. 

That's why I suggested renaming it to `ContainsAddr`, I believe you're checking if the `KernelSymbolTable` contains the address.

grantseltzer: I'll add that it's fine to have another helper function, but I think the name should be `KernelSymbol.IsInAddressRange` or something like that.

itamarmaouda101: Definitely Agree  (:

Your explain of the `KernelSymbolTable.IsInTextSegment()` and `KernelSymbol.ContainsAddr()` made me understand your idea, thanks! 

One little Q:

If ill rename it to be `KernelSymbolTable.ContainsAddr()', I will need to change the logic to only check for the address in the `kerenlSymbolTable` right?

So in that case ill have to create another function that check for address range , like you wrote about the `KernelSymbol.IsInAddressRange`

also I have the `GetSymbolByAddr` that check for symbol byt its address, isn't the same as the `containsAddr` idea?

rafaeldtinoco:
@itamarmaouda101 dont forget to:
```
git rebase -i origin/main
```
then change "pick" in front of your commit for "reword" and change the commit title and description to something meaningful to the entire change (not only your last action on it). Then you can force push again.

rafaeldtinoco (dismissed):
LGTM. @grantseltzer would you mind merging this if you are good with latest changes from Itamar ?

grantseltzer (approved):
LGTM!

PR #136, opened 2022-03-06
Fix the URL for tracee-ebpf

nguyenvulong:
I guess this dir has been moved to `/cmd`?


grantseltzer (approved):
Good catch, thank you!

PR #139, opened 2022-03-16
VersionString() API selftest

grantseltzer:
**FAQ**

Is this the most useful selftest? No. 

Is it good to have? ok.

### DIFF
--- a/selftest/version/main.go
+++ b/selftest/version/main.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/aquasecurity/libbpfgo"
+)
+
+func main() {
+
+	cmd := exec.Command("git", "describe", "--tags")
+	cmd.Dir = "../../libbpf"
+
+	b, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Fatal(string(b), err)
+	}
+
+	// libbpf doesn't put the patch version in exported version
+	// symbols, so use just prefix to exclude it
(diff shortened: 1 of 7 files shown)


kakkoyun (approved):

selftest/version/main.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import "C"
```
rafaeldtinoco, line 3: we dont need import C here.

selftest/version/main.go
```diff
@@ -0,0 +1,28 @@
...
+	cmd.Dir = "../../libbpf"
+
+	b, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// libbpf doesn't put the patch version in exported version
+	// symbols, so use just prefix to exclude it
+	if strings.HasPrefix(libbpfgo.VersionString(), string(b)) {
```
rafaeldtinoco, line 25: When I first read libbpfgo.VersionString() I thought we were comparing libbpf version with libbpfgo version (to see if libbpf string was contained in libbpfgo string). Should we change "VersionString" name ?

rafaeldtinoco (dismissed):
LGTM but I would remove the unneeded import and think about changing the version function name (it looks like we are giving libbpfgo version string and not libbpf one).

selftest/version/main.go
grantseltzer: Sure!

grantseltzer:
> 

Done!

rafaeldtinoco (approved):
LGTM

PR #153, opened 2022-04-19
fix kallsyms in libbpfgo helpers

itamarmaouda101:
kallsyms helper is including `[]` chars when it shouldn't.

This PR solves it by using TrimPrefix and TrimSurfix.


grantseltzer (approved):
L G T M !

PR #154, opened 2022-04-21
Add helpers for passing flags

grantseltzer:
The bpf helper `bpf_map_update_elem` takes a flags parameter which we're not letting users specify. Instead of breaking the API, i've added another helper for it. Additionally I've added API for the `bpf_map_lookup_elem_flags` helper.


kakkoyun (dismissed):
LGTM :100:

libbpfgo.go
```diff
@@ -841,6 +861,14 @@ func (b *BPFMap) Update(key, value unsafe.Pointer) error {
 	return nil
 }
 
+func (b *BPFMap) UpdateFlags(key, value unsafe.Pointer, flags MapFlag) error {
```
rafaeldtinoco, line 864: The function name makes it look like it is updating the "flags" for a map, while actually we're just updating the map using flags argument. Should we change the method name ?

I think I'm ok with not breaking the API to have flags added, at least until we decide to break the API and unify those 2 (should we add a TODO at the code ?). Also, we could make Update() be a call to UpdateFlags() (with a new name) with a nil flag argument (something like libbpf does when deprecating functions).

rafaeldtinoco (changes requested):
Code is good, I would re-think the function name, add a TODO and call one function from the other (unless you have better ideas, all ears).

grantseltzer:
Good thinking, I've updated accordingly.

rafaeldtinoco (approved):
LGTM. Make sure to "squashing" instead of "merging" (so we have a single commit). Thanks for addressing the suggestions.

grantseltzer:
I hate merging so much, that was an accident. It'l never happen again, I promise you!

PR #155, opened 2022-04-21
Spinlock selftest

grantseltzer:
This adds a selftest for spinlock functionality. This uses it for updating a counter in lost event logic for ringbuffers, but I don't think this is the best approach for that usecase. I plan on doing a writeup on alternatives.  

### DIFF
--- a/selftest/spinlocks/README.md
+++ b/selftest/spinlocks/README.md
@@ -1,0 +1,12 @@
+# Spinlocks
+
+https://lwn.net/Articles/776909/
+
+In BTF annotated BPF maps, we can use a bpf_spin_lock to allow for atomic reads/updates of values in BPF maps.
+
+It is NOT supported in programs of type:
+
+- BPF_PROG_TYPE_KPROBE
+- BPF_PROG_TYPE_TRACEPOINT
+- BPF_PROG_TYPE_PERF_EVENT
+- BPF_PROG_TYPE_RAW_TRACEPOINT
--- a/selftest/spinlocks/main.go
+++ b/selftest/spinlocks/main.go
@@ -1,0 +1,120 @@
+package main
+
+import "C"
+
+import (
+	"os"
+	"syscall"
+	"time"
+	"unsafe"
+
+	"encoding/binary"
+	"fmt"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+func resizeMap(module *bpf.Module, name string, size uint32) error {
+	m, err := module.GetMap("events")
+	if err != nil {
+		return err
+	}
+
+	if err = m.Resize(size); err != nil {
+		return err
+	}
+
+	if actual := m.GetMaxEntries(); actual != size {
+		return fmt.Errorf("map resize failed, expected %v, actual %v", size, actual)
+	}
+
+	return nil
+}
+
+func main() {
+
+	bpfModule, err := bpf.NewModuleFromFile("main.bpf.o")
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+	defer bpfModule.Close()
+
+	if err = resizeMap(bpfModule, "events", 8192); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+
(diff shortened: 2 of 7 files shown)


selftest/spinlocks/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+https://lwn.net/Articles/776909/
+
+In BTF annotated BPF maps, we can use a bpf_spin_lock to allow for atomic reads/updates of values in BPF maps.
+
+It is NOT supported in programs of type:
+
+- BPF_PROG_TYPE_KPROBE
+- BPF_PROG_TYPE_TRACEPOINT
+- BPF_PROG_TYPE_PERF_EVENT
```
kakkoyun, line 11: It's not mentioned in the attached article, do you know why it is not supported in certain types of programs? Is it because with these events, the atomicity is already guaranteed? I'd appreciate if you have a pointer to the related documentation.

grantseltzer: I'm not entirely sure why, but you can see [here](https://elixir.bootlin.com/linux/latest/source/tools/testing/selftests/bpf/verifier/helper_restricted.c#L126) where I got this from.

According to the [patch](https://patchwork.kernel.org/project/netdevbpf/patch/20211113142227.566439-3-me@ubique.spb.ru/) that added those verifier tests, these program types can result in some locking issues.

kakkoyun: Thanks for the pointers. This is very useful.

selftest/spinlocks/main.go
```diff
@@ -0,0 +1,120 @@
...
+		os.Exit(-1)
+	}
+
+	err = bpfModule.BPFLoadObject()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+
+	bpfModule.ListProgramNames()
```
rafaeldtinoco, line 54: image

We should not have a printf within libbpfgo C code, should we ? Maybe this is a leftover from the development ?

selftest/spinlocks/main.go
```diff
@@ -0,0 +1,120 @@
...
+	var lostEventCounterKey uint32 = 1
+	var zero uint32
+	lostEventCounterMap.Update(unsafe.Pointer(&lostEventCounterKey), unsafe.Pointer(&zero))
+
+	rb.Start()
+
+	numberOfEventsReceived := 0
+	go func() {
+		for {
+			syscall.Mmap(999, 999, 999, 1, 1)
```
rafaeldtinoco, line 93: Is this even working for you ? I added an error check:

image

and I'm getting:

image

and error from the syscall.

rafaeldtinoco: I wonder why you picked those arguments for the mmap syscall, any particular reason ? Would an anonymous memory mmap (at a random place) be enough ?

rafaeldtinoco (changes requested):
Logic looks ok but there is an issue with mmap call (at least for me). Maybe we should change the call to mmap and use something else ? Using something that could actually cause ring buffer reservations failure would be good.

Also, what if we reduce the size of the ringbuffer to force errors for a repetitive probe ? I think we have to improve this one (despite checking if the functions work).

selftest/spinlocks/main.go
grantseltzer: It's just to trigger the bpf code, the actual mmap call is irrelevant.

rafaeldtinoco: Ah ok, =D makes sense then.

grantseltzer:
> Logic looks ok but there is an issue with mmap call (at least for me). Maybe we should change the call to mmap and use something else ? Using something that could actually cause ring buffer reservations failure would be good.
> 
> Also, what if we reduce the size of the ringbuffer to force errors for a repetitive probe ? I think we have to improve this one (despite checking if the functions work).

Agreed that it would be useful to have this as a demonstration of what spinlocks could be useful for, but it's not necessarily the purpose of the selftest to be interacting with ringbuffers. I can do it anyway, and you're right that reducing the size would be the perfect way of doing so!

rafaeldtinoco (approved):

PR #156, opened 2022-04-27
Per CPU map needed API and selftest

grantseltzer:


libbpfgo.go
```diff
@@ -630,6 +638,15 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
 	return value, nil
 }
 
+func (b *BPFMap) GetValue2(key unsafe.Pointer, value *[]byte) error {
```
kakkoyun, line 641: Is it possible to have a better name than `GetValue2`? And also adding a documentation to explain the need of the additional method would be awesome!

grantseltzer: Hahaha yes, I forgot that I pushed it with that name. I'll update it, thanks for your reviews @kakkoyun !

libbpfgo.go
```diff
@@ -611,6 +611,14 @@ func (b *BPFMap) ValueSize() int {
 	return int(C.bpf_map__value_size(b.bpfMap))
 }
 
+func (b *BPFMap) SetValueSize(size uint32) error {
+	errC := C.bpf_map__set_value_size(b.bpfMap, C.uint(size))
```
javierhonduco, line 615: Perhaps we could read both the return value and the error, as reading just one of the return values will result in us reading the C function return value, but not the errno

libbpfgo.go
```diff
@@ -630,6 +638,15 @@ func (b *BPFMap) GetValue(key unsafe.Pointer) ([]byte, error) {
 	return value, nil
 }
 
+func (b *BPFMap) GetValue2(key unsafe.Pointer, value *[]byte) error {
+	valuePtr := unsafe.Pointer(&(*value)[0])
+	errC := C.bpf_map_lookup_elem(b.fd, key, valuePtr)
```
javierhonduco, line 643: Same as above

grantseltzer:
libbpf just merged [high level apis](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/diff/tools/lib/bpf/libbpf.h?id=737d0646a83c) for accessing maps, which includes built in API for reading from per-cpu maps. This should be changed over to use that API (though it hasn't been released yet)

selftest/percpu/main.go
```diff
@@ -0,0 +1,74 @@
...
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+	if link.GetFd() == 0 {
+		os.Exit(-1)
+	}
+
+	go func() {
+		for {
+			syscall.Mmap(999, 999, 999, 1, 1)
```
rafaeldtinoco, line 49: This call is broken (like the other selftest with the same call). If you check for error here you will see this is not working (at least in the other exact same call for another PR).

libbpfgo.go
grantseltzer: bpf_map__set_value_size() returns the error directly and does not set errno.

rafaeldtinoco (approved):
LGTM. Make sure to squash needed commits and dont duplicate "Signed off by" lines in git log! Thanks!

PR #158, opened 2022-04-27
Fixes for arm64

javierhonduco:
On my dev machine:
```
$ uname -m
arm64
```

This PR addresses this and creates a `GO` Make variable to make it easy to
override while trying golang binaries in non-standard paths

**Test plan**
```
$ sudo make GO=/usr/local/go/bin/go run-dynamic
[...]
sudo ./run.sh main-dynamic
[*] SUCCESS: all good
```

cc/ @Sylfrena @v-thakkar @kakkoyun

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,9 @@ SELFTEST = ./selftest
 
 CC = gcc
 CLANG = clang
+GO = go
 
-ARCH := $(shell uname -m)
-ARCH := $(subst x86_64,amd64,$(ARCH))
+ARCH := $(shell uname -m | sed 's/x86_64/amd64/g; s/aarch64/arm64/g')
 
 BTFFILE = /sys/kernel/btf/vmlinux
 BPFTOOL = $(shell which bpftool || /bin/false)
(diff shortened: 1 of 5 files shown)


kakkoyun (dismissed):
LGTM :rocket:

selftest/tcpconnect/Makefile
```diff
@@ -7,8 +7,11 @@ LIBBPF_OBJ = $(abspath $(OUTPUT)/libbpf.a)
 
 CC = gcc
 CLANG = clang
+GO = go
 
 ARCH = x86
```
grantseltzer, line 12: Is this line needed anymore?

Makefile
```diff
@@ -70,7 +70,7 @@ libbpfgo-static: $(VMLINUXH) | $(LIBBPF_OBJ)
 		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
 		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
 		GOOS=linux GOARCH=$(ARCH) \
```
grantseltzer, line 72: This change breaks the go build on x86 because the correct architecture argument for GOARCH is `amd64`. 

https://github.com/golang/go/blob/master/src/go/build/syslist.go#L52

grantseltzer (changes requested):

Makefile
javierhonduco: I am confused, I thought that it should work
```
$ echo x86_64 | sed 's/x86_64/amd64/g; s/aarch64/arm64/g'
amd64
```

What's the error you are getting?

selftest/tcpconnect/Makefile
javierhonduco: It's a convenience variable for folks that have Go installed outside of the PATH, e.g. when testing bleeding-edge Golang binaries with libbpfgo, but can drop it if you prefer :)

javierhonduco:
Removed a leftover arch=x86 variable

Makefile
grantseltzer: `cmd/go: unsupported GOOS/GOARCH pair linux/x86`. in `selftest/common/Makefile` youre replacing `x86_64` with `x86`, not `amd64`

javierhonduco: Oh! My bad. Was not looking at the right file!! 🤦

selftest/tcpconnect/Makefile
grantseltzer: I was referring to `ARCH = x86` in this case

Issue #159, opened 2022-04-28
Error handling improvements


javierhonduco:
👋 . Been using libbpfgo recently and found some problems with the way errors are being handled. Would love to hear your opinion and comments on this subject 😄 .

## Context

While helping @kakkoyun debug https://github.com/aquasecurity/libbpfgo/issues/147 to make our map retrieval and deletion path more efficient, we were puzzled as the kernel did not seem to return `-EPERM` in the code path we were executing.

This can be seen with this [bpftrace script](https://github.com/javierhonduco/bpf-playground/blob/master/bpfsnoop_delete_batch.sh):

```
$ sudo ./bpfsnoop_delete_batch.sh `pidof parca-agent`
Attaching 4 probes...

return from BPF syscall
        syscall+40
        0xfffffffff000
        runtime.asmcgocall.abi0+124
 with -2
return from libbpf
        _cgo_3f570fc81fe2_C2func_bpf_map_lookup_and_delete_batch+1196
        runtime.asmcgocall.abi0+124
 with -1
```

As we can see here, the "raw" (without any massaging of the return code by libc) BPF syscall is returning `-2` (`ENOENT`), while libbpf is seeing `-1` (`EPERM`). Why is this happening?

libc's `syscall(2)`, which libbpf calls [here](https://github.com/libbpf/libbpf/blob/3a4e26307d0f9b227e3ebd28b443a1a715e4e17d/src/bpf.c#L75) changes the return code to `-1` on error, and sets `errno` to the actual return value of the "raw" syscall.

The [`syscall(2)`](https://man7.org/linux/man-pages/man2/syscall.2.html) manpage:
> The return value is defined by the system call being invoked.  In
>        general, a 0 return value indicates success.  A -1 return value
>       indicates an error, and an error number is stored in [errno](https://man7.org/linux/man-pages/man3/errno.3.html).

🤔  So it seems we thought we were getting `EPERM` when in reality, it was "just" `-1` indicating that the function failed (*) and that we had to check `errno` to see what the actual error is. Interpreting the return value as `errno` is not correct here.

As far as I understand, cgo calls have two return values, the return code, and `errno` (https://pkg.go.dev/cmd/cgo). In several of libbpfgo's calls to libbpf, only one return value is checked, which if I am not mistaken, it would be the return code of the function.

## How this is affecting libbpfs APIs right now
Besides the batch APIs not returning the data correctly, even though we might have processed it just fine, this is also affecting some other APIs.

[...]

PR #160, opened 2022-04-28
Add strict mode API

javierhonduco:
See https://github.com/libbpf/libbpf/wiki/Libbpf-1.0-migration-guide for context.
Related to https://github.com/aquasecurity/libbpfgo/issues/159

cc/ @Sylfrena @kakkoyun @v-thakkar

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -328,6 +328,43 @@ func NewModuleFromFile(bpfObjPath string) (*Module, error) {
 	})
 }
 
+// LibbpfStrictMode is an enum as defined in https://github.com/libbpf/libbpf/blob/2cd2d03f63242c048a896179398c68d2dbefe3d6/src/libbpf_legacy.h#L23
+type LibbpfStrictMode uint32
+
+const (
+	LibbpfStrictModeAll               LibbpfStrictMode = C.LIBBPF_STRICT_ALL
+	LibbpfStrictModeNone              LibbpfStrictMode = C.LIBBPF_STRICT_NONE
+	LibbpfStrictModeCleanPtrs         LibbpfStrictMode = C.LIBBPF_STRICT_CLEAN_PTRS
+	LibbpfStrictModeDirectErrs        LibbpfStrictMode = C.LIBBPF_STRICT_DIRECT_ERRS
+	LibbpfStrictModeSecName           LibbpfStrictMode = C.LIBBPF_STRICT_SEC_NAME
+	LibbpfStrictModeNoObjectList      LibbpfStrictMode = C.LIBBPF_STRICT_NO_OBJECT_LIST
+	LibbpfStrictModeAutoRlimitMemlock LibbpfStrictMode = C.LIBBPF_STRICT_AUTO_RLIMIT_MEMLOCK
+	LibbpfStrictModeMapDefinitions    LibbpfStrictMode = C.LIBBPF_STRICT_MAP_DEFINITIONS
+)
+
+func (b LibbpfStrictMode) String() (str string) {
+	x := map[LibbpfStrictMode]string{
+		LibbpfStrictModeAll:               "LIBBPF_STRICT_ALL",
+		LibbpfStrictModeNone:              "LIBBPF_STRICT_NONE",
+		LibbpfStrictModeCleanPtrs:         "LIBBPF_STRICT_CLEAN_PTRS",
+		LibbpfStrictModeDirectErrs:        "LIBBPF_STRICT_DIRECT_ERRS",
+		LibbpfStrictModeSecName:           "LIBBPF_STRICT_SEC_NAME",
+		LibbpfStrictModeNoObjectList:      "LIBBPF_STRICT_NO_OBJECT_LIST",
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -328,6 +328,24 @@ func NewModuleFromFile(bpfObjPath string) (*Module, error) {
 	})
 }
 
+func SetStrictMode(mode string) error {
```
grantseltzer, line 331: Let's define a constant for this, similar to how we do for map and program types.

grantseltzer (changes requested):
Just the one comment, this is a very welcome addition, thank you very much!

javierhonduco:
Addressed review feedback + added some more strict mode values

libbpfgo.go
```diff
@@ -787,7 +822,7 @@ func (b *BPFMap) DeleteKeyBatch(keys unsafe.Pointer, count uint32) error {
 		ElemFlags: C.BPF_ANY,
 		Flags:     C.BPF_ANY,
 	}
-	
+
```
javierhonduco, line 825: Gofmt really wants this newline :D

libbpfgo.go
```diff
@@ -328,6 +328,41 @@ func NewModuleFromFile(bpfObjPath string) (*Module, error) {
...
+type LibbpfStrictMode uint32
+
+const (
+	LibbpfStrictModeAll               LibbpfStrictMode = C.LIBBPF_STRICT_ALL
+	LibbpfStrictModeNone              LibbpfStrictMode = C.LIBBPF_STRICT_NONE
+	LibbpfStrictModeCleanPtrs         LibbpfStrictMode = C.LIBBPF_STRICT_CLEAN_PTRS
+	LibbpfStrictModeDirectErrs        LibbpfStrictMode = C.LIBBPF_STRICT_DIRECT_ERRS
+	LibbpfStrictModeSecName           LibbpfStrictMode = C.LIBBPF_STRICT_SEC_NAME
+	LibbpfStrictModeNoObjectList      LibbpfStrictMode = C.LIBBPF_STRICT_NO_OBJECT_LIST
+	LibbpfStrictModeAutoRlimitMemlock LibbpfStrictMode = C.LIBBPF_STRICT_AUTO_RLIMIT_MEMLOCK
```
grantseltzer, line 341: We're missing `LIBBPF_STRICT_MAP_DEFINITIONS` here. I believe it was added in libbpf v0.7.0

libbpfgo.go
```diff
@@ -328,6 +328,41 @@ func NewModuleFromFile(bpfObjPath string) (*Module, error) {
 	})
 }
 
+// LibbpfStrictMode is an enum as defined in https://elixir.bootlin.com/linux/v5.17.5/source/tools/lib/bpf/libbpf_legacy.h#L23
```
grantseltzer, line 331: Let's link to libbpf explicitly, instead of the linux source code which isn't typically in sync (libbpf changes make their way down into mainline)

https://github.com/libbpf/libbpf/blob/2cd2d03f63242c048a896179398c68d2dbefe3d6/src/libbpf_legacy.h#L23

libbpfgo.go
grantseltzer: Gofmt is so needy

grantseltzer (changes requested):

libbpfgo.go
javierhonduco: Oh, good call, thought it was the other way around for some reason!

javierhonduco:
Addressed feedback

libbpfgo.go
```diff
@@ -328,6 +328,43 @@ func NewModuleFromFile(bpfObjPath string) (*Module, error) {
...
+
+	str, ok := x[b]
+	if !ok {
+		str = LibbpfStrictModeNone.String()
+	}
+	return str
+}
+
+func SetStrictMode(mode LibbpfStrictMode) {
+	C.libbpf_set_strict_mode(uint32(mode))
```
grantseltzer, line 365:
```suggestion
	C.libbpf_set_strict_mode(C.uint(mode))
```

grantseltzer: Last thing, then good to merge :+1:

javierhonduco: Adding this errors with:
```
../libbpfgo/libbpfgo.go:365:27: cannot use _Ctype_uint(mode) (value of type _Ctype_uint) as type uint32 in argument to (_Cfunc_libbpf_set_strict_mode)
```

Seems like `uint32` is the right type 😄. Passing a "Go" type seems recurrent for enums. I tried searching for this behaviour, but my Google-fu failed me

https://github.com/aquasecurity/libbpfgo/blob/main/libbpfgo.go#L1158-L1160

grantseltzer: That's so strange, I can in fact reproduce that. 

I did some googling and apparently enums are simplified to uint32's in CGO wrappers. [This issue](https://github.com/golang/go/issues/40494) isn't the same as ours specifically, but it does mention this.

grantseltzer (approved):
L G T M !

Issue #162, opened 2022-05-03
Ensure that batch APIs handle partial success conditions


javierhonduco:
More context on https://github.com/aquasecurity/libbpfgo/issues/159.

- [ ] `GetValueAndDeleteBatch`
- [ ] `DeleteKeyBatch`
- [ ] `GetValueBatch`
- [ ] `UpdateBatch`

geyslan:
`UpdateBatch` is a special case. We cannot ignore the error of updating fewer entries than were requested.

Thinking about the Batch API, I suppose we should break it by returning the `count` updated by the kernel, thus leaving it up to the consumer to decide if it's a failure or not.

```diff
diff --git a/map-low.go b/map-low.go
index 7cf3d95..fd6ed73 100644
--- a/map-low.go
+++ b/map-low.go
@@ -247,7 +247,7 @@ func (m *BPFMapLow) DeleteKey(key unsafe.Pointer) error {
 // BPFMapLow Batch Operations
 //
 
-func (m *BPFMapLow) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
+func (m *BPFMapLow) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, uint32, error) {
        valueSize, err := calcMapValueSize(m.ValueSize(), m.Type())
        if err != nil {
                return nil, fmt.Errorf("map %s %w", m.Name(), err)
@@ -291,7 +291,7 @@ func (m *BPFMapLow) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.
        return collectBatchValues(values, uint32(countC), valueSize), nil
 }
 
-func (m *BPFMapLow) GetValueAndDeleteBatch(keys, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
+func (m *BPFMapLow) GetValueAndDeleteBatch(keys, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, uint32, error) {
        valueSize, err := calcMapValueSize(m.ValueSize(), m.Type())
        if err != nil {
                return nil, fmt.Errorf("map %s %w", m.Name(), err)
@@ -328,7 +328,7 @@ func (m *BPFMapLow) GetValueAndDeleteBatch(keys, startKey, nextKey unsafe.Pointe
[...]

Issue #147, opened 2022-04-12
"operation not permitted" errors for batch operations


kakkoyun:
I have been trying to debug the reason for this, but so far I haven't managed to be successful. Thus, I'm asking for help.

We try to use `GetValuesAndBatch` and it receives "operation not permitted" error.

https://github.com/parca-dev/parca-agent/blob/bd9807a3a0e16302b5944d570967ef5a828dfc80/pkg/profiler/profiler.go#L344-L358

I have tried to bump the `rlimits` (usually that's the culprit under this error) but no luck with that either
https://github.com/parca-dev/parca-agent/blob/bd9807a3a0e16302b5944d570967ef5a828dfc80/pkg/profiler/profiler.go#L623-L642 

Do you happen to have any pointers or guideline for me to further debug this? Or could this be related to error handling?

grantseltzer:
Hm, i'll investigate this now. I typically run with tracee in background to see what syscall is giving the EPERM and go from there.

grantseltzer:
@kakkoyun Do you have an issue on the parca agent describing this so we can discuss how to reproduce?

grantseltzer:
I'm seeing that if the `count` parameter passed to `bpf_map_lookup_and_delete_batch` is greater than the number of elements in the map that I get EPERM.

Checking out the [documentation](https://libbpf.readthedocs.io/en/latest/api.html) for the libbpf function, count is an input and output parameter, so you should be able to print that value to see if anything is being read/deleted before the permission denied error occurs.

kakkoyun:
@grantseltzer I think we were passing the capacity of the map, and we need to pass the size of the map, (still tying to validate)

In any case, from the documentation I understand that even though AP returns error it could be a partial success. And the index of the last successful operation is indicated by the count in/out parameter. Is this also want you understand from it?

If it's the case, the current batch APIs don't consider this fact.

grantseltzer:
> I assume you debugged it using tracee, right? I want to add that to our debugging flow if it's case :)

I would have but no I have no tried running parca-agent yet. I do recommend tracee for debugging though! It's easier to use for debugging than strace.

> In any case, from the documentation I understand that even though AP returns error it could be a partial success. And the index of the last successful operation is indicated by the count in/out parameter. Is this also want you understand from it?

Yes that's what I understand as well (I wrote that documentation btw haha)

> If it's the case, the current batch APIs don't consider this fact.

[...]

PR #163, opened 2022-05-04
Fix some more error handling

javierhonduco:
This PR doesn't address all the call-sites that need to have their return error checked but focuses on some APIs that I am experimenting with. For more context, see: https://github.com/aquasecurity/libbpfgo/issues/159

(this PR does not tackle the remaining batch API calls https://github.com/aquasecurity/libbpfgo/issues/162)

Thanks!

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -566,30 +566,30 @@ func (b *BPFMap) SetType(mapType MapType) error {
 
 func (b *BPFMap) Pin(pinPath string) error {
 	path := C.CString(pinPath)
-	errC := C.bpf_map__pin(b.bpfMap, path)
+	ret, errC := C.bpf_map__pin(b.bpfMap, path)
 	C.free(unsafe.Pointer(path))
-	if errC != 0 {
-		return fmt.Errorf("failed to pin map %s to path %s: %w", b.name, pinPath, syscall.Errno(-errC))
+	if ret != 0 {
+		return fmt.Errorf("failed to pin map %s to path %s: %w", b.name, pinPath, errC)
 	}
 	return nil
 }
 
 func (b *BPFMap) Unpin(pinPath string) error {
 	path := C.CString(pinPath)
-	errC := C.bpf_map__unpin(b.bpfMap, path)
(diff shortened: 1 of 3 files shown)


libbpfgo.go
```diff
@@ -600,9 +600,9 @@ func (b *BPFMap) SetPinPath(pinPath string) error {
 // Note: for ring buffer and perf buffer, maxEntries is the
 // capacity in bytes.
 func (b *BPFMap) Resize(maxEntries uint32) error {
-	errC := C.bpf_map__set_max_entries(b.bpfMap, C.uint(maxEntries))
-	if errC != 0 {
-		return fmt.Errorf("failed to resize map %s to %v: %w", b.name, maxEntries, syscall.Errno(-errC))
+	ret, errC := C.bpf_map__set_max_entries(b.bpfMap, C.uint(maxEntries))
+	if ret != 0 {
+		return fmt.Errorf("failed to resize map %s to %v: %w", b.name, maxEntries, errC)
```
grantseltzer, line 605: Why don't we want to use `syscall.Errno(-errC)`? In these cases you're using %w for errC, which is an int?

javierhonduco: The second argument's type of any CGO call is `error`, which is the errno already.

It's confusing because Go doesn't require us to bind all the return values of a function, and before, as `errC` was the sole function return value we were reading, it was in reality the return value, which is a `C.int`, and not what we want.

That's why passing it to `syscall.Errno` always resulted in `-1`, `EPERM`(https://github.com/aquasecurity/libbpfgo/issues/147) [1]


[1]: For completeness, it returns `-1`on failure before libbpf 1.0, which hasn't been released yet, or if `LibbpfStrictModeDirectErrs` is not used (https://github.com/libbpf/libbpf/wiki/Libbpf-1.0-migration-guide).

Hope this makes sense, let me know if this is not what you meant!

grantseltzer: Ah yes of course, my confusion. I'd like to standardize naming of CGO errors that are `int`'s as `errorCode` and the second return of type error as `errno`. I can handle that in the future.

grantseltzer (approved):
LGTM!

PR #168, opened 2022-05-16
PoC of checking available kernel features to determine what bpf programs to load

grantseltzer:
See `samples/feature-detect/README.md`

Note that the intention isn't to merge this into libbpfgo as is, I would remove reference to tracee before that, but it would be good to have as a sample. This is just a convenient way of discussing the PoC.


grantseltzer:
Closing this in favor of https://github.com/aquasecurity/tracee/pull/1846

PR #183, opened 2022-06-27
Remove ListProgramNames API

grantseltzer:
This API function relies on a `printf()` call in C code, which I was using for debugging purposes. It shouldn't be included in the upcoming release so I'm removing it. 


rafaeldtinoco (approved):
LGTM

Issue #192, opened 2022-07-21
add parsing for bpf program type


roikol:
this parsing helper function would be used by the Tracee PR implementing this issue: https://github.com/aquasecurity/tracee/issues/2003

rafaeldtinoco:
I believe the issue should be more descriptive (and use the template) about the need of the "helper function".

Im still not sure if the interface and methods existing in libbpfgo (already created to parse ebpf program types, and other stuff, are enough for your need).

PR #193, opened 2022-07-21
parse bpf program type

roikol:
solve: #192

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -31,16 +31,3 @@ jobs:
       - name: Static Selftests
         run: |
           sudo make selftest-static-run
-  compile-trace:
-    name: Compile Tracee
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v2
-      - name: Install Dependencies
-        uses: ./.github/actions/build-dependencies
-        with:
-          go-version: 1.18
-      - name: Compile Tracee
-        run: |
-          sudo ./testing/compile-tracee.sh
--- a/.github/workflows/test-daily.yaml
+++ b/.github/workflows/test-daily.yaml
@@ -1,0 +1,24 @@
+name: Tests (DAILY)
+on:
+  workflow_dispatch: {}
+  schedule:
+    # Daily at 03:00
+    - cron: '0 3 * * *'
+  pull_request:
+    branches:
+      - main
+  workflow_call:
+jobs:
+  compile-trace:
+    name: Compile Tracee
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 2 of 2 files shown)


helpers/argumentParsers.go
```diff
@@ -1766,3 +1766,133 @@ func ParseSocketOption(rawValue uint64) (SocketOptionArgument, error) {
 	}
 	return v, nil
 }
+
+type BpfProgTypeArgument int32
```
rafaeldtinoco, line 1770: I believe we have a way to get program type added in libbpfgo already:

https://github.com/aquasecurity/libbpfgo/blob/216bd5b702c8c8867d2a431f7050740b24c3a152/libbpfgo.go#L1127-L1128

wouldn't that interface be enough ?

rafaeldtinoco, line 1770: What is this different from the type BPFProgType already existent ? Can you re-use the type and functions we already have for bpf prog type ?

rafaeldtinoco: If you check line 1128 and on, you will find we have bpf prog type, bpf map type, bpf attachment types, etc.

roikol: thanks @rafaeldtinoco,

i didn't see that.

the `BPFProgType` is defined in `package libbpfgo`. 
i think that the parsing function i'm suggesting should be in `package helpers`. so, using this already existing type would create a problem of circular dependencies - `package helpers` would import `package libbpfgo` which imports `package helpers`.

how do you suggest i can use this existing type? 
move it to a new package maybe? or move it into `package helpers`?

i agree we shouldn't define same type with its funcs more than once.

grantseltzer: @roikol I think that helpers should be able to import libbpfgo but libbpfgo should _not_ import helpers. Before this PR it was only imported for the RWArray (line 1504). 

I think RWArray should be moved over to the libbpfgo package, or to it's own package. That way you don't have to redefine all the same constants again here, and you can cleanly have the parse functions in helpers.

I will open a PR momentarily with this fix, once that's merged you can update this PR.

roikol: thank you @grantseltzer !

sounds good. wasn't really sure what's the best approach here.
so i'll wait for your PR to be merged and then i'll update this one.

thanks :)

grantseltzer: Take a look at #202  when you get a chance

grantseltzer: Merged 202, you can rebase your changes on top of that

roikol:
@grantseltzer 
accidently pushed 'main' branch into this one, so it merged this PR automatically (with 0 commits).
opened a new one instead with the correct changes #209 .

PR #194, opened 2022-07-23
Add selftest and run scripts

imhunterand:
Signed-off-by: redact<redact@redac.com>


rafaeldtinoco:
Not sure what this is about, but looks very wrong.

PR #197, opened 2022-07-31
fix: check KernelSymbolTable initialized before getting symbols

AsafEitani:
Add checks before accessing the kernel symbol table


rafaeldtinoco (approved):
LGTM

PR #198, opened 2022-08-03
batch: Fix DeleteKeyBatch

javierhonduco:
when dealing with partial deletion, more context on https://github.com/aquasecurity/libbpfgo/issues/162

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -880,7 +880,8 @@ func (b *BPFMap) UpdateBatch(keys, values unsafe.Pointer, count uint32) error {
 // DeleteKeyBatch allows for batch deletion of multiple elements in the map.
 //
 // `count` number of keys will be deleted from the map. Passing an argument that greater than the number of keys
-// in the map will cause the function to return a syscall.EPERM as an error.
+// in the map will cause the function to delete fewer keys than requested. See the inline comment in
+// `GetValueAndDeleteBatch` for more context.
 func (b *BPFMap) DeleteKeyBatch(keys unsafe.Pointer, count uint32) error {
 	countC := C.uint(count)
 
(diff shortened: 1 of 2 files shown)


javierhonduco:
Updated docs

libbpfgo.go
```diff
@@ -890,15 +891,11 @@ func (b *BPFMap) DeleteKeyBatch(keys unsafe.Pointer, count uint32) error {
...
-		if sc != syscall.EFAULT {
-			if uint32(countC) != count {
-				return fmt.Errorf("failed to batch delete ALL keys from map %s, deleted (%d/%d): %w", b.name, uint32(countC), count, sc)
-			}
-		}
-		return fmt.Errorf("failed to batch delete keys from map %s: %w", b.name, syscall.Errno(-errC))
+	ret, errC := C.bpf_map_delete_batch(b.fd, keys, &countC, bpfMapBatchOptsToC(opts))
+
+	if ret != 0 && errC != syscall.ENOENT {
+		// ret = -1 && errno == syscall.ENOENT indicates a partial read and delete.
```
kakkoyun, line 897: In this case, is it a partial failure or complete? Can you clarify the comment a little more?

javierhonduco: In this condition, there's a "partial success", we have processed some of the elements, but not as many as requested.

Happy to add more context here. Did not want to duplicate the big comment in `GetValueAndDeleteBatch`, that's why I referred to it in this method's doc

javierhonduco: No, the `ret = -1 && errno == syscall.ENOENT` means there was a partial success and in that case we don't error out. We don't error either in case of a full success.

Only when there's a real error (`ret != 0 && errC != syscall.ENOENT`) is when we return an error to the user

kakkoyun: Yes, as you can tell, when that comment is in the condition scope, my parser throws an exception :see_no_evil: Maybe you should write the above before the `if` for elderly people such as me :older_man:

javierhonduco: Ohh good point, the placement is not amazing I agree. Changing it :)

kakkoyun (approved):

grantseltzer (approved):
LGTM! Thanks so much for the contributions @javierhonduco !

PR #199, opened 2022-08-03
batch: Fix GetValueBatch

javierhonduco:
when dealing with partial gets, more context on #162

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -742,11 +742,11 @@ func bpfMapBatchOptsToC(batchOpts *BPFMapBatchOpts) *C.struct_bpf_map_batch_opts
 // Once the first iteration is complete you can provide the last key seen in the previous iteration as the startKey for the next iteration
 // and repeat until nextKey is nil.
 //
-// The last argument, count, is the number of keys to lookup. Passing an argument that greater than the number of keys
-// in the map will cause the function to return a syscall.EPERM as an error.
+// The last argument, count, is the number of keys to lookup. The kernel will update it with the count of the elements that were
+// retrieved.
 //
-// The API can return partial results even though an error is returned.
-// In this case the keys that were successfully retrieved until an error occurred will be in the result slice.
+// The API can return partial results even though an -1 is returned. In this case, errno will be set to `ENOENT` and the values slice and count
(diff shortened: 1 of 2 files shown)


javierhonduco:
Updated the docs for the function

libbpfgo.go
```diff
@@ -760,19 +760,16 @@ func (b *BPFMap) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Poi
...
-				return collectBatchValues(values, uint32(countC), b.ValueSize()),
-					fmt.Errorf("failed to retrieve ALL elements in map %s, fetched (%d/%d): %w", b.name, uint32(countC), count, sc)
-			}
-		}
-		return nil, fmt.Errorf("failed to batch lookup values %v in map %s: %w", keys, b.name, syscall.Errno(-errC))
+	ret, errC := C.bpf_map_lookup_batch(b.fd, startKey, nextKey, keys, valuesPtr, &countC, bpfMapBatchOptsToC(opts))
+	processed := uint32(countC)
+
+	if ret != 0 && errC != syscall.ENOENT {
+		// ret = -1 && errno == syscall.ENOENT indicates a partial read.
```
kakkoyun, line 767: In this case, is it a partial failure or complete? Can you clarify the comment a little more?

javierhonduco: In this condition, there's a "partial success", we have processed some of the elements, but not as many as requested.

Happy to add more context here. Did not want to duplicate the big comment in `GetValueAndDeleteBatch`, that's why I referred to it in this method's doc

javierhonduco: We do that below with `return collectBatchValues(values, processed, b.ValueSize())`. Either with a partial or a complete success we return there.

What I meant before is that `ret = -1 && errno == syscall.ENOENT` means partial success and we won't error in this case, we will return the subset of the elements that got read. This usually means that the requested number of elements was higher than the existing number of elements. 

Otherwise, we return with an error in line 768, and this means that there is a legit error and nothing got processed

kakkoyun (approved):

grantseltzer (approved):
LGTM!

Issue #149, opened 2022-04-13
Run all selftests on each PR w/ Github actions


grantseltzer:

kakkoyun:
Big +1 to this!

Issue #37, opened 2021-07-12
Run selftests in github actions


grantseltzer:

rafaeldtinoco:
@danielpacak this is the issue for running the selftests in actions (that you looked into in our talk today). FYIO

PR #178, opened 2022-06-23
Update PR workflow to trigger on PRs

grantseltzer:
Signed-off-by: grantseltzer <grantseltzer@gmail.com>

PR #201, opened 2022-08-08
Update PR workflow to trigger on PRs

mozillazg:
https://github.com/aquasecurity/libbpfgo/pull/178

closes https://github.com/aquasecurity/libbpfgo/issues/149
closes https://github.com/aquasecurity/libbpfgo/issues/37

Result: https://github.com/mozillazg/libbpfgo/pull/1

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -10,17 +10,12 @@ jobs:
     name: unit test helpers
     runs-on: ubuntu-20.04
     steps:
-      - name: Setup Go
-        uses: actions/setup-go@v1
-        with:
-          go-version: 1.18
       - name: Checkout Code
         uses: actions/checkout@v2
-      - name: install dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install --yes zlib1g-dev libelf-dev
-        shell: bash
+      - name: Install Dependencies
+        uses: ./.github/actions/build-dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,73 @@
...
+          sudo apt-get install --yes build-essential
+          sudo apt-get install --yes pkgconf
+          sudo apt-get install --yes llvm-12 clang-12
+          sudo apt-get install --yes clang-format-12
+          sudo apt-get install --yes zlib1g-dev libelf-dev
+          for tool in "clang" "llc" "llvm-strip" "clang-format"
+          do
+            sudo rm -f /usr/bin/$tool
+            sudo ln -s /usr/bin/$tool-12 /usr/bin/$tool
+          done
```
rafaeldtinoco, line 69: We should use `build-dependencies` action file here (like https://github.com/aquasecurity/tracee/blob/main/.github/actions/build-dependencies/action.yaml) IMO.

rafaeldtinoco:
image

This test is expected to fail after recent merges. That is actually something useful from this test: we know when tracee has been broken by changes in libbpfgo as well. Thanks!

.github/workflows/pr.yaml
mozillazg: done!

PR #202, opened 2022-08-08
Move RWArray helpers into own package

grantseltzer:
The RWArray is the only part of the helpers package
that was previous used within the libbpfgo package.
This presents an issue when the helpers package wants
to import libbpfgo (creates a circular dependency)

Moving RWArray to a different package solves this issue.

In general, helpers should be able to import libbpfgo but
libbpfgo should not import helpers

### DIFF
--- a/helpers/rwarray/rwArray.go
+++ b/helpers/rwarray/rwArray.go
@@ -1,4 +1,4 @@
-package helpers
+package rwarray
 
 import (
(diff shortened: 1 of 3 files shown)


roikol:
@grantseltzer  LGTM!

i've also tested your branch with my changes regarding #193, and it works good - no circular dependency issue

PR #203, opened 2022-08-09
argparser: divide socket option parsing to get and set

AlonZivony:
There is a single const value which has different meaning according to
the handling function - SO_GET_FILTER in `getsockopt` which is
SO_ATTACH_FILTER in `setsockopt`.
Divided the parsing of socket option to be according to origin
function, so the arg name received will be precise.
Also removed unused socket option with dup value.


rafaeldtinoco (approved):
LGTM

PR #196, opened 2022-07-24
Libbpfgo cgroup attach

rafaeldtinoco:
commit e47fe84
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Sun Jul 24 01:26:42 2022

    BPFProg: create AttachCgroup to allow Cgroup link attachments

    Create AttachCgroup(string) to allow the following program types:

    - BPF_PROG_TYPE_CGROUP_SKB
    - BPF_PROG_TYPE_CGROUP_SOCK
    - BPF_PROG_TYPE_CGROUP_DEVICE
    - BPF_PROG_TYPE_CGROUP_SOCK_ADDR
    - BPF_PROG_TYPE_CGROUP_SYSCTL
    - BPF_PROG_TYPE_CGROUP_SOCKOPT

    to be attached to a cgroupv2 directory file descriptor, such as:

    - /sys/fs/cgroup/unified
    - /sys/fs/cgroup/unified/user.slice/user-1000.slice

    and on...

    Those eBPF program types will be triggered for all processes within the
    cgroupv2 they were attached to.

commit f531670
Author: Rafael David Tinoco <rafaeldtinoco@gmail.com>
Date:   Sat Jul 23 23:57:36 2022

    BPFProg: move AttachLSM to a correct place

    - Keep doAttach...() function together

PR #204, opened 2022-08-09
Add selftest for cgroup attach

mozillazg:
https://github.com/aquasecurity/libbpfgo/pull/196


mozillazg:
selftest:

image

selftest/cgroup/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+	err = bpfModule.BPFLoadObject()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+
+	cgroupRootDir := "/sys/fs/cgroup/unified"
+	if _, err := os.Stat(cgroupRootDir); os.IsNotExist(err) {
+		cgroupRootDir = "/sys/fs/cgroup"
```
rafaeldtinoco, line 31: We have to check whether this is a cgroupv2 directory or not (cgroup v1 is not supported). Best way to do this is to check for the "release_agent" file existence iirc (cgroup v2 doesn't have it).

selftest/cgroup/main.go
```diff
@@ -0,0 +1,81 @@
...
+	rb, err := bpfModule.InitRingBuf("events", eventsChannel)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(-1)
+	}
+
+	rb.Start()
+	numberOfEventsReceived := 0
+	go func() {
+		_, err := exec.Command("ping", "localhost", "-c 10").Output()
```
rafaeldtinoco, line 59: This will create a single socket and test needs 5 socket creations, right ? We should spawn `ping localhost -c 1 -w 1` for 5 times instead ? Or something like it...

selftest/cgroup/main.go
geyslan: https://man7.org/linux/man-pages/man7/cgroups.7.html

```
Note that on many modern systems, [systemd(1)](https://man7.org/linux/man-pages/man1/systemd.1.html) automatically mounts
       the cgroup2 filesystem at /sys/fs/cgroup/unified during the boot
       process.
```

This is not so true, since in my arch based it's mounted at /sys/fs/cgroup. So checking whether it's unified or not would not help in this case.

```bash
$ mount | grep cgroup
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime)

$ mount | grep cgroup | cut -d " " -f 1
cgroup2
```

Perhaps checking whether `/sys/fs/cgroup/cgroup.controllers` is available then it would be v2?

```
The new behaviors in cgroups v2 are summarized here, and in some
       cases elaborated in the following subsections.

       1. Cgroups v2 provides a unified hierarchy against which all
          controllers are mounted.
...
       3. Active cgroups must be specified via the files
          cgroup.controllers and cgroup.subtree_control.
```

Or checking as systemd does: https://github.com/systemd/systemd/blob/5c19169fe95f436a233f92fc62fdca783a0fc5ec/src/basic/cgroup-util.c#L2099

It's summarized here: https://unix.stackexchange.com/a/617824/47040

geyslan: @rafaeldtinoco what main distros do we need to keep in mind?

rafaeldtinoco: The thing is cgroupv2 support might be available even if cgroupv2 is not mounted. Distros might choose to mount cgroupv1, cgroupv1+v2 or cgroupv2 only. The way we solved that in tracee is to temporarily mount a cgroupv2 filesystem if not mounted, check availability and use that directory if needed.

Systemd provides a way to choose if you want to have mixed cgroup v1 and cgroup v2 mounted filesystems (systemd.unified_cgroup_hierarchy=XXX). For this test I would say checking if cgroupv2 is mounted (through /proc/mounts and check "cgroup2") would be enough (and to use that dir).

selftest/cgroup/main.go
mozillazg: fixed

selftest/cgroup/main.go
mozillazg: fixed

geyslan: @mozillazg Thank you for the time into this.

selftest/cgroup/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"regexp"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+var reCgroup2Mount = regexp.MustCompile(`(?m)^cgroup2\s+(\S+)\s+`)
```
geyslan, line 16: Perhaps `(?m)^cgroup2\s\/[\S]+\scgroup2\s`, catching mount point start `/` and the type as `cgroup2`. WDYT?

```suggestion
var reCgroup2Mount = regexp.MustCompile(`(?m)^cgroup2\s\/[\S]+\scgroup2\s`)
```

selftest/cgroup/main.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func getCgroupV2RootDir() string {
+	data, err := ioutil.ReadFile("/proc/mounts")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "read /proc/mounts failed: %+v\n", err)
+		os.Exit(-1)
+	}
+	items := reCgroup2Mount.FindStringSubmatch(string(data))
+	if len(items) < 2 {
```
geyslan, line 91: Following suggestion above:

```suggestion
	cgroup2Match := reCgroup2Mount.FindString(string(data))
	if len(cgroup2Match) == 0 {
```

selftest/cgroup/main.go
```diff
@@ -0,0 +1,96 @@
...
+		fmt.Fprintln(os.Stderr, "cgroupv2 is not mounted")
+		os.Exit(-1)
+	}
+	return items[1]
```
geyslan, line 95: Following suggestion above:

```suggestion
	cgroup2Match := reCgroup2Mount.FindString(string(data))
	if len(cgroup2Match) == 0 {
		fmt.Fprintln(os.Stderr, "cgroupv2 is not mounted")
		os.Exit(-1)
	}
	return strings.Split(cgroup2Match, " ")[1]
```

selftest/cgroup/main.go
geyslan: Or, going with your `FindStringSubmatch` approach, just set the regex to `(?m)^cgroup2\s\/(\S+)\scgroup2\s`. Be my guest.

mozillazg: Yes. Your way is better.

mozillazg: updated

geyslan (approved):
LGTM

mozillazg:
I have no idea how this happened :
https://github.com/aquasecurity/libbpfgo/runs/7815708108?check_suite_focus=true
image

Maybe we should use `make` instead of `make -j8` in Makefile?:
https://github.com/aquasecurity/libbpfgo/blob/0850b8cf46a9610acbe6b2c8503fb69a88af4243/Makefile#L129-L138

rafaeldtinoco:
> Maybe we should use `make` instead of `make -j8` in Makefile?:
> 
> https://github.com/aquasecurity/libbpfgo/blob/0850b8cf46a9610acbe6b2c8503fb69a88af4243/Makefile#L129-L138

Hey @mozillazg, the "-j8" was a way to "accelerate" the libbpf builds by default in most of the envs (even when no number of jobs was informed). If it is causing issue we can remove it for sure (and rely in -j argument).

PR #209, opened 2022-08-11
parse bpf program type

roikol:
add helper to parse bpf program type

### DIFF
--- a/helpers/argumentParsers.go
+++ b/helpers/argumentParsers.go
@@ -7,6 +7,8 @@ import (
 	"net"
 	"strconv"
 	"strings"
+
+	"github.com/aquasecurity/libbpfgo"
 )
 
 type SystemFunctionArgument interface {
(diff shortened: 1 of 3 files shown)


rafaeldtinoco:
@geyslan (To check the unit tests error regarding use of inexistent OS dynamic library, instead of using the static lib we build).

helpers/argumentParsers.go
```diff
@@ -1705,3 +1707,7 @@ func ParseGetSocketOption(rawValue uint64) (SocketOptionArgument, error) {
 	}
 	return v, nil
 }
+
+func ParseBpfProgType(rawValue uint64) (libbpfgo.BPFProgType, error) {
+	return libbpfgo.BPFProgType(rawValue), nil
```
grantseltzer, line 1712: We should have a check here to make sure that this is a valid BPF program type. For example, `ParseSocketDomainArgument` has a map of known types. We could implement something similar in libbpfgo.go. What do you think?

grantseltzer (changes requested):

helpers/argumentParsers.go
roikol: i agree. it's more correct to have this check here.
in the `String()` function of `libbpfgo.BPFProgType` we have something like that, but the helper parsing function should return an error if the value is not a valid BPF program type.

i'll push an update soon

libbpfgo.go
```diff
@@ -1201,6 +1203,41 @@ func (b BPFProgType) String() (str string) {
 	return str
 }
 
+var BpfProgTypeMap = map[uint64]BPFProgType{
```
grantseltzer, line 1206: I think this map should live in `helpers` next to the ParseBPFProgType() function, this way this map doesn't have to be exported.

helpers/argumentParsers_test.go
```diff
@@ -147,3 +148,37 @@ func TestParseGetSocketOption(t *testing.T) {
 		})
 	}
 }
+
+func TestParseBpfProgType(t *testing.T) {
```
grantseltzer, line 152:
```suggestion
func TestParseBPFProgType(t *testing.T) {
```

helpers/argumentParsers.go
```diff
@@ -1705,3 +1707,11 @@ func ParseGetSocketOption(rawValue uint64) (SocketOptionArgument, error) {
 	}
 	return v, nil
 }
+
+func ParseBpfProgType(rawValue uint64) (libbpfgo.BPFProgType, error) {
```
grantseltzer, line 1711:
```suggestion
func ParseBPFProgType(rawValue uint64) (libbpfgo.BPFProgType, error) {
```

geyslan:
> @geyslan (To check the unit tests error regarding use of inexistent OS dynamic library, instead of using the static lib we build).

https://github.com/aquasecurity/libbpfgo/pull/212 merged.

grantseltzer (approved):
LGTM!

PR #213, opened 2022-08-17
Add APIs to utilize map of maps

kakkoyun:
Add selftest for the map of map interactions.

## TODO

* [ ] Properly test (I'm having problems testing this on the ARM64)

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -523,7 +523,7 @@ func bpfMapCreateOptsToC(createOpts *BPFMapCreateOpts) *C.struct_bpf_map_create_
 // and therefore do not have access to libbpf high level bpf_map__* APIS
 // which causes different behavior from maps created in the kernel side code
 //
-// See usage of `bpf_map_create()` in kernel selftests for more info
+// See usage of `bpf_map_create()` in kernel selftests for more info.
 func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries int, opts *BPFMapCreateOpts) (*BPFMap, error) {
 	cs := C.CString(mapName)
 	fdOrError := C.bpf_map_create(uint32(mapType), cs, C.uint(keySize), C.uint(valueSize), C.uint(maxEntries), bpfMapCreateOptsToC(opts))
@@ -534,6 +534,27 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
 
 	return &BPFMap{
 		name:   mapName,
+		fd:     fdOrError,
+		module: nil,
+		bpfMap: nil,
+	}, nil
+}
+
+// GetMapByID returns a BPF map by ID. This can be used for reaching the maps
+// from a BPF array of maps or hash of maps. However, this function uses a low-level
+// libbpf API; maps loaded in this way do not conform to libbpf map formats,
+// and therefore do not have access to libbpf high level bpf_map__* APIS
+// which causes different behavior from maps created in the kernel side code.
+//
+// See usage of `bpf_map_get_fd_by_id()` in kernel selftests for more info.
+func (m *Module) GetMapByID(id uint32) (*BPFMap, error) {
+	fdOrError := C.bpf_map_get_fd_by_id(C.uint(id))
+	if fdOrError < 0 {
+		return nil, fmt.Errorf("could not find map: %w", syscall.Errno(-fdOrError))
+	}
+
+	return &BPFMap{
+		name:   "unknown",
 		fd:     fdOrError,
 		module: nil,
 		bpfMap: nil,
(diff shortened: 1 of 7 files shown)


selftest/map-of-maps/main.go
```diff
@@ -0,0 +1,124 @@
...
+	"os"
+	"syscall"
+	"unsafe"
+
+	"github.com/aquasecurity/libbpfgo"
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+/*
+	Interactions with hash of maps and array of map uses `CreateMao` and `GetMapByID`` which use
```
javierhonduco, line 18:
```suggestion
	Interactions with map of maps and array of map uses `CreateMap` and `GetMapByID` which use
```

libbpfgo.go
```diff
@@ -540,6 +540,27 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
...
 
+// GetMapByID returns a BPF map by ID. This can be used for reaching the maps
+// from a BPF array of maps or hash of maps. However, this function uses a low-level
+// libbpf API; maps loaded in this way do not conform to libbpf map formats,
+// and therefore do not have access to libbpf high level bpf_map__* APIS
+// which causes different behavior from maps created in the kernel side code.
+//
+// See usage of `bpf_map_get_fd_by_id()` in kernel selftests for more info.
+func (m *Module) GetMapByID(id uint32) (*BPFMap, error) {
+	fdOrError := C.bpf_map_get_fd_by_id(C.uint(id))
```
javierhonduco, line 551: Should we perhaps check the second return value (`errno`)?

libbpfgo.go
```diff
@@ -610,6 +631,17 @@ func (b *BPFMap) SetPinPath(pinPath string) error {
 	return nil
 }
 
+// SetInnerMap is used to set the inner map of a hash of map or an array of map.
+// This is only to satify the eBPF verifier and needs to be called before loading the module.
+func (b *BPFMap) SetInnerMap(innerMap *BPFMap) error {
+	// b is the outer map.
+	errC := C.bpf_map__set_inner_map_fd(b.bpfMap, C.int(innerMap.fd))
```
javierhonduco, line 638: Same as above

libbpfgo.go
```diff
@@ -924,17 +956,21 @@ func (b *BPFMap) DeleteKey(key unsafe.Pointer) error {
...
 func (b *BPFMap) Update(key, value unsafe.Pointer) error {
 	return b.UpdateValueFlags(key, value, MapFlagUpdateAny)
 }
 
 func (b *BPFMap) UpdateValueFlags(key, value unsafe.Pointer, flags MapFlag) error {
+	var val interface{} = (*BPFMap)(value)
+	m, ok := val.(*BPFMap)
+	if ok {
+		value = unsafe.Pointer(&m.fd)
+	}
```
javierhonduco, line 973: Should we handle the case when the type wasn't of `BPFMap`?

javierhonduco (commented):
Left some initial comments, will do another pass later

libbpfgo.go
kakkoyun: In that case, we can just past the value as is. We only have a special case if the given value is a map.

I'm not sure if it's the best way to check a type from an unsafe pointer. If you know any better way let me know.

javierhonduco: I see, not really a thing we should be concerned about here. Anyhow, perhaps something like `(*(*BPFMap)(value))` would be clearer?

kakkoyun: It's still not an assertion though. It's casting. But I can still use it while assigning the interface. This could be a stupid effort, in any case I'll test if further. 

It sucks that the api doesn't accept bpf_map rather than a fd.

kakkoyun: This was a rather stupid attempt. Type casting panics if the underlying type doesn't match.

kakkoyun:
> Hi @kakkoyun Should this still be set as a draft? Do you think it's ready for review and if not can I help you with it?

This is still a draft. I'll give it another shot now that I'm back to my usual working environment (better for testing), hopefully soon.

I could use an initial review and suggestions regarding the APIs though.

grantseltzer:
@kakkoyun Sounds good! i'll take a look later today. There may be a new API in libbpf 1.0 that could help with this as well.

geyslan:
@kakkoyun do you still have plans to get going with this PR? Thanks for contributing.

kakkoyun:
@geyslan I still want to have it in libbpfgo. I just need to spare time to figure out how to make it happen. Feel free to help me guide me :)

geyslan:
> @geyslan I still want to have it in libbpfgo. I just need to spare time to figure out how to make it happen. Feel free to help me guide me :)

Ok then. First, could you resolve the conflicts? After call me for review. 👍🏼

rafaeldtinoco:
Is there any interest in continuing this work ? I'm afraid we did not have enough time to rebase and fix the issues, I'll keep it opened for sometime to see if there is any movement here. Cheers!

kakkoyun:
> Is there any interest in continuing this work ? I'm afraid we did not have enough time to rebase and fix the issues, I'll keep it opened for sometime to see if there is any movement here. Cheers!

I would love to revisit this. However, I don't have enough free-cycles to work on this. Please feel free to just close it. I'll get back to it when I have time again.

geyslan:
@kakkoyun I'll continue this. 👍🏼

geyslan:
I gave some air to this in #343.

PR #214, opened 2022-08-22
legacy cgroup attachment functionality (needed for <= 5.6 kernels)

rafaeldtinoco:
I'm still working on this. Will finish the selftest and try to REAL use cases before moving away from draft. But feel free to review the work so far. Thanks.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -5,7 +5,9 @@ package libbpfgo
 
 #include <errno.h>
 #include <stdlib.h>
+#include <unistd.h>
 #include <sys/resource.h>
+#include <sys/syscall.h>
 
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
@@ -30,26 +32,40 @@ static inline long PTR_ERR(const void *ptr)
 }
 #endif
 
-int libbpf_print_fn(enum libbpf_print_level level, const char *format,
+int libbpf_print_fn(enum libbpf_print_level level,
+					const char *format,
                     va_list args)
 {
-    if (level != LIBBPF_WARN)
-        return 0;
+	if (level != LIBBPF_WARN)
+		return 0;
+
+	va_list check;
+	va_copy(check, args);
+	char *str = va_arg(check, char *);
 
 	// BUG: https://github.com/aquasecurity/tracee/issues/1676
 
-	va_list check; va_copy(check, args);
-	char *str = va_arg(check, char *);
 	if (strstr(str, "Exclusivity flag on") != NULL) {
 		va_end(check);
 		return 0;
 	}
-	va_end(check);
 
-    return vfprintf(stderr, format, args);
+	// AttachCgroupLegacy() will first try AttachCgroup() and it
+	// might fail. This is not an error and is the best way of
+	// probing for eBPF cgroup attachment link existence.
+
+	str = va_arg(check, char *);
+	if (strstr(str, "cgroup") != NULL) {
+		str = va_arg(check, char *);
+		if (strstr(str, "Invalid argument") != NULL)
+			return 0;
+	}
+
+	return vfprintf(stderr, format, args);
 }
 
-void set_print_fn() {
+void set_print_fn()
+{
     libbpf_set_print(libbpf_print_fn);
 }
 
@@ -85,6 +101,35 @@ struct perf_buffer * init_perf_buf(int map_fd, int page_cnt, uintptr_t ctx)
     }
 
     return pb;
+}
+
+int bpf_prog_attach_cgroup_legacy(
+	int prog_fd,          // eBPF program file descriptor
+	int target_fd,        // cgroup directory file descriptor
+	int type)             // BPF_CGROUP_INET_{INGRESS,EGRESS}, ...
+{
+	union bpf_attr attr;
+	memset(&attr, 0, sizeof(attr));
+	attr.target_fd = target_fd;
+	attr.attach_bpf_fd = prog_fd;
+	attr.attach_type = type;
+	attr.attach_flags = BPF_F_ALLOW_MULTI; // or BPF_F_ALLOW_OVERRIDE
+
+	return syscall(__NR_bpf, BPF_PROG_ATTACH, &attr, sizeof(attr));
+}
+
+int bpf_prog_detach_cgroup_legacy(
+	int prog_fd,          // eBPF program file descriptor
+	int target_fd,        // cgroup directory file descriptor
+	int type)             // BPF_CGROUP_INET_{INGRESS,EGRESS}, ...
+{
+	union bpf_attr attr;
+	memset(&attr, 0, sizeof(attr));
+	attr.target_fd = target_fd;
+	attr.attach_bpf_fd = prog_fd;
+	attr.attach_type = type;
+
+	return syscall(__NR_bpf, BPF_PROG_DETACH, &attr, sizeof(attr));
 }
 */
 import "C"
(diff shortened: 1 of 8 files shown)


geyslan:
LGTM so far. I can't run the tests on a generic/ubuntu2004 box, unfortunately.

```bash
vagrant@ubuntu2004:~/libbpfgo/selftest/cgroup-legacy$ uname -a
Linux ubuntu2004.localdomain 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
vagrant@ubuntu2004:~/libbpfgo/selftest/cgroup-legacy$ pwd
/home/vagrant/libbpfgo/selftest/cgroup-legacy
vagrant@ubuntu2004:~/libbpfgo/selftest/cgroup-legacy$ git branch
* cgroup-attach-legacy
  main
  tinoco
vagrant@ubuntu2004:~/libbpfgo/selftest/cgroup-legacy$ make
make -C /home/vagrant/libbpfgo libbpfgo-static
make[1]: Entering directory '/home/vagrant/libbpfgo'
CC=clang \
	CGO_CFLAGS="-I/home/vagrant/libbpfgo/output" \
	CGO_LDFLAGS="-lelf -lz /home/vagrant/libbpfgo/output/libbpf.a" \
	GOOS=linux GOARCH=amd64 \
	go build \
	-tags netgo -ldflags '-w -extldflags "-static"' \
	.
make[1]: Leaving directory '/home/vagrant/libbpfgo'
make -C /home/vagrant/libbpfgo vmlinuxh
make[1]: Entering directory '/home/vagrant/libbpfgo'
make[1]: Leaving directory '/home/vagrant/libbpfgo'
clang -g -O2 -Wall -fpie -target bpf -D__TARGET_ARCH_amd64 -I../../output -c main.bpf.c -o main.bpf.o
In file included from main.bpf.c:3:
In file included from ../../output/bpf/bpf_helpers.h:11:
../../output/bpf/bpf_helper_defs.h:73:83: error: unknown type name '__u64'
static long (*bpf_map_update_elem)(void *map, const void *key, const void *value, __u64 flags) = (void *) 2;
                                                                                  ^
../../output/bpf/bpf_helper_defs.h:97:42: error: unknown type name '__u32'
static long (*bpf_probe_read)(void *dst, __u32 size, const void *unsafe_ptr) = (void *) 4;
...
```

rafaeldtinoco:
> ```shell
> vagrant@ubuntu2004:~/libbpfgo/selftest/cgroup-legacy$ uname -a
> Linux ubuntu2004.localdomain 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

If this is a Focal machine, please check tracee/packaging/Dockerfile.ubuntu-packaging file to see how to properly install needed toolchain (clang-12 needed, etc).

libbpfgo.go
```diff
@@ -1331,6 +1384,61 @@ func (p *BPFProg) AttachCgroup(cgroupV2DirPath string) (*BPFLink, error) {
...
 
+// AttachCgroupLegacy attaches the BPFProg to a cgroup described by given fd.
+// It first tries to use the most recent attachment method and, if that does
+// not work, instead of failing it tries the legacy way: to attach the cgroup
+// eBPF program without previously creating a link. Related to upstream kernel
+// commit af6eea57437a ("bpf: Implement bpf_link-based cgroup BPF program
+// attachment"). This allows attaching cgroup eBPF ingress/egress in older
+// kernels.
+func (p *BPFProg) AttachCgroupLegacy(cgroupV2DirPath string, attachType BPFAttachType) (*BPFLink, error) {
+	// bpfLink, err := p.AttachCgroup(cgroupV2DirPath)
```
rafaeldtinoco, line 1395: I'll un-comment this in the final merge (it is like that so anyone testing this code will correctly test it in newer kernels as well). I'll check if 5.4 kernels are good before merging as well.

libbpfgo.go
```diff
@@ -239,23 +270,35 @@ const (
...
 type BPFLink struct {
-	link      *C.struct_bpf_link
-	prog      *BPFProg
-	linkType  LinkType
-	eventName string
+	link       *C.struct_bpf_link
+	prog       *BPFProg
+	linkType   LinkType
+	eventName  string
+	attachType BPFAttachType // needed by legacy cgroup attachment
```
grantseltzer, line 282: Do you think we should change the name from `attachType` to something like "attachTypeLegacy" to note that this instead of a comment? To make it less confusing to implement APIs that create BPFLinks. (same comment for cgroupDir)

Also this got me confused about the difference between link/attach types (which I have before and then forgot again). I opened https://github.com/aquasecurity/libbpfgo/issues/217, just noting.

libbpfgo.go
```diff
@@ -239,23 +270,35 @@ const (
...
+	case CgroupLegacy:
+		return l.prog.DetachCgroupLegacy(l.cgroupDir, l.attachType)
+	}
+	return fmt.Errorf("unable to destroy legacy link")
 }
 
 func (l *BPFLink) Destroy() error {
-	ret := C.bpf_link__destroy(l.link)
-	if ret < 0 {
+	if l.link == nil { // might be a legacy/fake BPFLink
```
grantseltzer, line 295: Should we explicitly mark links as being created via this legacy api so as to better surface errors?

libbpfgo.go
```diff
@@ -1331,6 +1384,61 @@ func (p *BPFProg) AttachCgroup(cgroupV2DirPath string) (*BPFLink, error) {
 	return bpfLink, nil
 }
 
+// AttachCgroupLegacy attaches the BPFProg to a cgroup described by given fd.
+// It first tries to use the most recent attachment method and, if that does
+// not work, instead of failing it tries the legacy way: to attach the cgroup
+// eBPF program without previously creating a link. Related to upstream kernel
+// commit af6eea57437a ("bpf: Implement bpf_link-based cgroup BPF program
+// attachment"). This allows attaching cgroup eBPF ingress/egress in older
+// kernels.
```
grantseltzer, line 1393:
```suggestion
// AttachCgroupLegacy attaches the BPFProg to a cgroup described by given fd.
// It first tries to use the most recent attachment method and, if that does
// not work, instead of failing it tries the legacy way; to attach the cgroup
// eBPF program without previously creating a link. 
// 
// This allows attaching cgroup eBPF ingress/egress in older kernels
//
// Related upstream kernel commit: https://github.com/torvalds/linux/commit/af6eea57437a

```

libbpfgo.go
```diff
@@ -1331,6 +1384,61 @@ func (p *BPFProg) AttachCgroup(cgroupV2DirPath string) (*BPFLink, error) {
...
+		cgroupDir: cgroupV2DirPath,
+	}
+	return fakeBpfLink, nil
+}
+
+// DetachCgroupLegacy detaches the BPFProg from a cgroup described by given fd.
+// This is needed because in legacy attachment there is no BPFLink, just a fake
+// one (kernel did not support it, nor libbpf). This function should be called
+// by the (*BPFLink)->Destroy() function, since BPFLink is emulated (so user
+// don´t need to distinguish between regular and legacy cgroup detachments).
```
grantseltzer, line 1427:
```suggestion
// DetachCgroupLegacy detaches the BPFProg from a cgroup described by the given fd.
// This is needed because in legacy attachment there is no BPFLink, just a fake
// one (kernel did not support it, nor libbpf). This function should be called
// by the (*BPFLink)->Destroy() function, since BPFLink is emulated (so users
// don´t need to distinguish between regular and legacy cgroup detachments).
```

libbpfgo.go
geyslan: > Also this got me confused about the difference between link/attach types (which I have before and then forgot again). I opened https://github.com/aquasecurity/libbpfgo/issues/217, just noting.

It makes sense. I would also use something like "attachTypeLegacy".

rafaeldtinoco: I see your point... but BPFAttachType is unrelated to the "legacy" concept as well.

Every BPFLink should have its "attachType" set (and it does inside libbpf) IF libbpf wasn't able to detect the needed attachment type automatically (there is a table that relates SEC(), eBPF program types and their attachment capabilities).

Libbpfgo BPFLink interface is unaware of their attachment types, when instantiated, because of the interface libbpf provides... but it might be needed to know which attachment type that link used (or even influence which to use if multiple available).

NEVERTHELESS, I believe you have a valid point by "isolating" what is legacy from what is not and I did a change:

```go
type BPFLinkLegacy struct {
	attachType BPFAttachType
	cgroupDir  string
}

type BPFLink struct {
	link      *C.struct_bpf_link
	prog      *BPFProg
	linkType  LinkType
	eventName string
	legacy    *BPFLinkLegacy
}
```

already addressing your concern bellow (about link not existing).

libbpfgo.go
rafaeldtinoco: Doing so in the version about to be pushed.

rafaeldtinoco:
The test still works in newer kernels but with a warning:

image

Because it is just testing the newer cgroup attachment type in those cases. When we run in kernels <= 5.4, then we get no warning and we test the interface for real. I thought about adding a flag to "force" using the legacy method but I don't think I should, so... it is what it is =).

grantseltzer (approved):
LGTM! I like what you did with those latest changes (selftest running regardless, legacy struct)

PR #215, opened 2022-08-23
makefile: fix helpers-test-static target

rafaeldtinoco:
Allow environment variables passing through sudo for the build target.


Makefile
```diff
@@ -175,7 +175,7 @@ helpers-test-static-run: libbpfgo-static
 	CC=$(CLANG) \
 		CGO_CFLAGS=$(CGO_CFLAGS_STATIC) \
 		CGO_LDFLAGS=$(CGO_LDFLAGS_STATIC) \
-		sudo $(GO) test -v $(HELPERS)/...
+		sudo -E $(GO) test -v $(HELPERS)/...
```
geyslan, line 178: Could it be the same for `helpers-test-dynamic-run`?

rafaeldtinoco: It was just a test. I want to clear the tests up, moved it back to 1.17 (as that is what we currently need to support).

PR #225, opened 2022-08-24
Standardize error handling in libbpfgo, remove rlimit memory bump

grantseltzer:
This standardizes error handling in libbpfgo to conform to libbpf 1.0 conventions which can be summarized as follows (from [Journey to libbpf 1.0](https://nakryiko.com/posts/libbpf-v1/):
```
any pointer-returning API returns NULL on error, and sets errno to actual (positive) Exxx error code;
any other API that might fail returns (negative) -Exxx error code directly as an int result and sets (positive) errno to Exxx.
```

So i've standardized in libbpfgo as follows:

Libbpf APIs that return an error code directly:

```
ret := C.libbpf_function()
if ret != 0 {
	return fmt.Errorf(“blah blah %w”, syscall.Errno(-ret))
}
```

libbpf APIs that return a pointer:
```
ptr, errno := C.libbpf_function()
if ptr == nil {
	return fmt.Errorf(“blah blah: %w”, errno)
}
```

**Additionally this PR removes bumping rlimit_memlock, libbpf does this automatically starting in libbpf 1.0**

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -12,28 +12,9 @@ package libbpfgo
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
 
-#ifndef MAX_ERRNO
-#define MAX_ERRNO           4095
-#define IS_ERR_VALUE(x)     ((x) >= (unsigned long)-MAX_ERRNO)
-
-static inline bool IS_ERR(const void *ptr)
-{
-    return IS_ERR_VALUE((unsigned long)ptr);
-}
-
-static inline bool IS_ERR_OR_NULL(const void *ptr)
-{
-    return !ptr || IS_ERR_VALUE((unsigned long)ptr);
-}
-
-static inline long PTR_ERR(const void *ptr)
-{
-    return (long) ptr;
-}
-#endif
 
 int libbpf_print_fn(enum libbpf_print_level level,
-					const char *format,
+                    const char *format,
                     va_list args)
 {
 	if (level != LIBBPF_WARN)
@@ -367,17 +348,6 @@ type RingBuffer struct {
 	wg     sync.WaitGroup
 }
 
-func errptrError(ptr unsafe.Pointer, format string, args ...interface{}) error {
-	negErrno := C.PTR_ERR(ptr)
-	errno := syscall.Errno(-int64(negErrno))
-	if errno == 0 {
-		return fmt.Errorf(format, args...)
-	}
-
-	args = append(args, errno.Error())
-	return fmt.Errorf(format+": %v", args...)
-}
-
 type NewModuleArgs struct {
 	KConfigFilePath string
 	BTFObjPath      string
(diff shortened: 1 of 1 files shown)


geyslan:
@grantseltzer based on the link provided `errno` will be positive and consistently set in both cases (API returning or not NULL). Whit that in hands what do you think about standardizing getting more closer to the go idiom?

```go
anyReturn, errno := C.libbpf_function()
if errno != 0 {
	return fmt.Errorf(“blah blah: %w”, syscall.Errno(errno))
}
```

A caveat is that I can't guarantee that the global errno is always cleared by the libbpf.

grantseltzer:
> @grantseltzer based on the link provided `errno` will be positive and consistently set in both cases (API returning or not NULL). Whit that in hands what do you think about standardizing getting more closer to the go idiom?
> 
> ```go
> anyReturn, errno := C.libbpf_function()
> if errno != 0 {
> 	return fmt.Errorf(“blah blah: %w”, syscall.Errno(errno))
> }
> ```

This is fair feedback. The historical context is that libbpf maintainers wanted to move away from setting errno in these type of functions and instead move towards directly returning error codes. They kept setting errno to keep backwards compatibility for the APIs that already were doing that. I don't think libbpf will remove setting errno, but directly checking is the 'intended' use. I guess it comes down to preference?

Also just to note, in your code snippet the `errno` second return is already of type error, don't need to cast it as a syscall.Errno.

> A caveat is that I can't guarantee that the global errno is always cleared by the libbpf.

I once questioned this and Go does something weird to reset errno for each go routine before CGO calls, it's a little sketchy lol

geyslan:
> I don't think libbpf will remove setting errno, but directly checking is the 'intended' use. I guess it comes down to preference?

Yeah. I got the context and agree.

> Also just to note, in your code snippet the errno second return is already of type error, don't need to cast it as a syscall.Errno.

My mistake. Then checking errno would be more go idiomatic, right?

```go
anyReturn, err := C.libbpf_function()
if err != nil {
	return fmt.Errorf(“blah blah: %w”, err)
}
```
Although I agree with your proposed standardization.

> I once questioned this and Go does something weird to reset errno for each go routine before CGO calls, it's a little sketchy lol

Interesting, indeed.

rafaeldtinoco (commented):
Could you create 2 commits, one for the errors other for the memory limit ? I want it clear in git history that we removed the memory limits raising (because of libbpf bump).

libbpfgo.go
```diff
@@ -533,7 +486,7 @@ func (m *Module) Close() {
 	}
 	for _, link := range m.links {
 		if link.link != nil {
-			C.bpf_link__destroy(link.link)
+			link.Destroy()
```
rafaeldtinoco, line 489: Reminder(myself): to check tcProbes and tracingProbes Destroy() for errors once we bump tracee.

rafaeldtinoco (commented):
It all LGTM if split into 2 commits AND it is much better we have a consistent way for error checking.

grantseltzer:
> Could you create 2 commits, one for the errors other for the memory limit ? I want it clear in git history that we removed the memory limits raising (because of libbpf bump).

You got it.

rafaeldtinoco (approved):
LGTM

PR #230, opened 2022-08-30
refact: standardize structure values access calls

### LINKED ISSUE #219: Do not use libbpf 'getter' aliases
libbpf standardized around a style of not using 'get' to describe functions that access values inside a struct. For example:

`bpf_program__type` instead of `bpf_program__get_type`. 

The latter will work but it's an alias of the former. We should use the correct form directly.

geyslan:
Closes #219

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -304,8 +304,13 @@ func (l *BPFLink) Destroy() error {
 	return nil
 }
 
-func (l *BPFLink) GetFd() int {
+func (l *BPFLink) FileDescriptor() int {
 	return int(C.bpf_link__fd(l.link))
+}
+
+// Deprecated: use BPFLink.FileDescriptor() instead.
+func (l *BPFLink) GetFd() int {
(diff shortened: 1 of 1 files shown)


geyslan:
@grantseltzer I wonder if we should, at this moment, add 1:1 naming like:

```go
func (b *BPFMap) PinPath() string {
	pinPath := C.GoString(C.bpf_map__pin_path(b.bpfMap))
	return pinPath
}
```

And make the old convention rely on that for retro-compatibility:

```go
func (b *BPFMap) GetPinPath() string {	
	return b.PinPath()
}
```

grantseltzer:
@geyslan I think that's a great idea, and we can remove the deprecated functions in libbpfgo 1.0

geyslan:
> @geyslan I think that's a great idea, and we can remove the deprecated functions in libbpfgo 1.0

So, soon I'll force push this again. Thanks.

Issue #228, opened 2022-08-29
CompareKernelRelease helper should correct format


grantseltzer:
The `CompareKernelRelease` helper function assumes that the passed kernel version formats are correct (i.e. `x.y.z`) which is unsafe because it can cause index out of bounds runtime errors, and is inconvenient in case something like `1.0` is passed.

grantseltzer:
Closed by #232

PR #232, opened 2022-08-31
Add possible error to helper functions for comparing kernel functions

grantseltzer:
#228 

### DIFF
--- a/helpers/common.go
+++ b/helpers/common.go
@@ -60,9 +60,23 @@ func UnameMachine() (string, error) {
 	return arch, nil
 }
 
+type KernelVersionComparison int
+
+const (
+	KernelVersionInvalid KernelVersionComparison = iota - 1
+	KernelVersionEqual
+	KernelVersionOlder
+	KernelVersionNewer
+)
+
 // CompareKernelRelease will compare two given kernel version/release
-// strings and return -1, 0 or 1 if given version is less, equal or bigger,
-// respectively, than the given one
+// strings and returns a KernelVersionComparison constant that shows
+// the comparsion of the given kernel version to the base.
+// For example CompareKernelRelease("5.8.1", "4.12.3") == KernelVersionOlder
+// It also returns an error incase of a malformed kernel version.
+//
+// Consumers should use the constants defined in this package for checking
+// the results: KernelVersionOlder, KernelVersionEqual, KernelVersionNewer
 //
 // Examples of $(uname -r):
 //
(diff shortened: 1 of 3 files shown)


helpers/common.go
```diff
@@ -72,26 +83,44 @@ func UnameMachine() (string, error) {
...
+func CompareKernelRelease(base, given string) (int, error) {
 	b := strings.Split(base, "-") // [base]-xxx
 	b = strings.Split(b[0], ".")  // [major][minor][patch]
 
+	if len(b) == 2 {
+		b = append(b, "0")
+	}
+	if len(b) != 3 {
+		return KernelVersionInvalid, fmt.Errorf("invalid base kernel version format: %s", base)
+	}
```
geyslan, line 95: I know it's not common but what if `base`/`given` are just "5". Could we interpret them as 5.0.0?

```suggestion
	b := strings.Split(base, "-") // [base]-xxx
	b = strings.Split(b[0], ".")  // [major][minor][patch]
	
	if len(b) == 0 {
		return KernelVersionInvalid, fmt.Errorf("invalid base kernel version format: %s", base)
	}

	for len(b) < 3 {
		b = append(b, "0")
	}
```

WDYT?

helpers/common.go
```diff
@@ -72,26 +83,44 @@ func UnameMachine() (string, error) {
...
+
 	g := strings.Split(given, "-")
 	g = strings.Split(g[0], ".")
+	if len(g) == 2 {
```
geyslan, line 99: ditto

helpers/common.go
```diff
@@ -72,26 +83,44 @@ func UnameMachine() (string, error) {
...
+		return KernelVersionInvalid, fmt.Errorf("invalid given kernel version format: %s", given)
+	}
 
 	for n := 0; n <= 2; n++ {
-		i, _ := strconv.Atoi(g[n])
-		j, _ := strconv.Atoi(b[n])
+		i, err := strconv.Atoi(g[n])
+		if err != nil {
+			return KernelVersionInvalid, fmt.Errorf("invalid given kernel version value: %s", g[n])
+		}
```
geyslan, line 110: Perhaps returning `given` plus `g[n]` together with strconv error? It would provide more information for the consumer.

grantseltzer:
@geyslan I really liked your suggestion, I implemented them and added tests

helpers/common.go
```diff
@@ -60,9 +60,23 @@ func UnameMachine() (string, error) {
 	return arch, nil
 }
 
+type KernelVersionComparison int
+
+const (
+	KernelVersionOlder   KernelVersionComparison = -1
+	KernelVersionEqual                           = 0
+	KernelVersionNewer                           = 1
+	KernelVersionInvalid                         = -3
```
geyslan, line 69: I think we need to repeat the type here since we have assignments, otherwise consts but first will be `const ... untyped int`.

If the const values are not required for outside use, another way can be using iota:

```suggestion
	KernelVersionOlder KernelVersionComparison = iota
	KernelVersionEqual
	KernelVersionNewer
	KernelVersionInvalid
```

geyslan (approved):

geyslan:
LGTM

PR #234, opened 2022-09-01
Clarify code in comments and var names

grantseltzer:
These functions are little confusing so this should help


rafaeldtinoco (approved):
LGTM

PR #237, opened 2022-09-08
builder: vagrant: introduce Vagrantfile

geyslan:
The contributor can make use of a VM box (arm64/amd64) using the following Makefile rules: `vagrant-up`, `vagrant-ssh`, `vagrant-halt` and `vagrant-destroy`.

One can override the ARCH env var: `make vagrant-up ARCH=arm64`

Whether the override will work depends on the availability of online boxes and compatibility of architecture and providers (vbox, parallels).

Tested in Darwin (M1) and Linux (amd64).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,10 @@ HELPERS = ./helpers
 CC = gcc
 CLANG = clang
 GO = go
+VAGRANT = vagrant
 
-ARCH := $(shell uname -m | sed 's/x86_64/amd64/g; s/aarch64/arm64/g')
+HOSTOS = $(shell uname)
+ARCH ?= $(shell uname -m | sed 's/x86_64/amd64/g; s/aarch64/arm64/g')
 
 BTFFILE = /sys/kernel/btf/vmlinux
 BPFTOOL = $(shell which bpftool || /bin/false)
@@ -179,6 +181,26 @@ helpers-test-static-run: libbpfgo-static
 
 helpers-test-dynamic-run: libbpfgo-dynamic
 	sudo $(GO) test -v $(HELPERS)/...
+
+# vagrant
+
+VAGRANT_DIR = $(abspath ./builder)
+
+.PHONY: vagrant-up
+.PHONY: vagrant-destroy
+.PHONY: vagrant-halt
+.PHONY: vagrant-ssh
+
+vagrant-up: .vagrant-up
+vagrant-destroy: .vagrant-destroy
+vagrant-halt: .vagrant-halt
+vagrant-ssh: .vagrant-ssh
+
+.vagrant-%:
+	VAGRANT_VAGRANTFILE=$(VAGRANT_DIR)/Vagrantfile-ubuntu \
+		ARCH=$(ARCH) \
+		HOSTOS=$(HOSTOS) \
+		$(VAGRANT) $*
 
 # output
 
--- a/builder/prepare-ubuntu.sh
+++ b/builder/prepare-ubuntu.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+#
+# This shell script is meant to prepare a building/exec environment for libbpfgo.
+#
+
+
+# variables
+
+[ -z "${GO_VERSION}" ] && GO_VERSION="1.17"
+[ -z "${CLANG_VERSION}" ] && CLANG_VERSION="12"
+[ -z "${ARCH}" ] && ARCH="amd64"
+
+
+# functions
+
+die() {
+    echo "ERROR: ${*}"
+    exit 1
+}
+
+info() {
+    echo "INFO: ${*}"
+}
+
+check_tooling() {
+    local tools="sudo apt-get"
+    for tool in ${tools}
+    do
+        command -v "${tool}" >/dev/null 2>&1 || die "missing required tool ${tool}"
+    done
+}
+
+install_pkgs() {
+    # silence 'dpkg-preconfigure: unable to re-open stdin: No such file or directory'
+    export DEBIAN_FRONTEND=noninteractive
+
+    sudo -E apt-get update || die "coud not update package list"
+    for pkg in "${@}"
+    do
+        info "Installing ${pkg}"
+        sudo -E apt-get install -y "${pkg}" || die "could not install ${pkg}"
+        info "${pkg} installed"
+    done
+}
+
+setup_go() {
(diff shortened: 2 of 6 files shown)


geyslan:
I put the environment building logic inside `builder/prepare-ubuntu.sh` so it can be reused by .github workflows, by instance.

rafaeldtinoco (commented):
If this change is for a contributor, we should create a markdown file explaining how to use it, giving examples of outputs and on how to obtain a shell and do the needed compilations IMO.

geyslan:
> If this change is for a contributor, we should create a markdown file explaining how to use it, giving examples of outputs and on how to obtain a shell and do the needed compilations IMO.

For sure, I'm on this as well. I spent these days running some builds and I realized today that this current proposal has the libbpfgo build broken due to the old version of libbpf and bpftool available in ubuntu 20.04. I was discussing with @grantseltzer that if we should stick with the bpftool requirement, it would be best to compile libbpf and bpftool altogether. BTW I'm changing the Vagrantfile to use 22.04 boxes.

builder/prepare-ubuntu.sh
```diff
@@ -0,0 +1,81 @@
...
+CMDS="wget rm tar chown apt-get"
+for cmd in ${CMDS}; do
+    command -v "${cmd}" >/dev/null 2>&1 || die "missing required tool ${cmd}"
+done
+
+install_go() {
+    info "Installing Go ${GOVERSION}"
+
+    local file="go${GOVERSION}.linux-${ARCH}.tar.gz"
+    local link="https://golang.org/dl/${file}"
```
rafaeldtinoco, line 33: Ubuntu has all golang versions supported (integrated to the OS) and patched (golang-1.17-go, golang-1.18-go, golang-1.19-go) and 1 is the default (golang-go). Can't you rely on the default repository ? Then you can simply "update-alternatives" to /usr/bin/go and /usr/bin/gopls to the version you want (dpkg -L golang-1.17-go | grep bin).

builder/prepare-ubuntu.sh
```diff
@@ -0,0 +1,81 @@
...
+info() {
+    echo "INFO: ${*}"
+}
+
+USER="${1}"
+[ -z "${USER}" ] && USER="root"
+USERHOME=$(eval echo ~"${USER}")
+[ -d "${USERHOME}" ] || die "directory ${USERHOME} doesn't exist"
+
+[ -z "${GOVERSION}" ] && GOVERSION="1.17.13"
```
rafaeldtinoco, line 21: ditto

builder/prepare-ubuntu.sh
```diff
@@ -0,0 +1,81 @@
...
+done
+
+install_go() {
+    info "Installing Go ${GOVERSION}"
+
+    local file="go${GOVERSION}.linux-${ARCH}.tar.gz"
+    local link="https://golang.org/dl/${file}"
+    sudo wget --progress=bar:force "${link}" -P /tmp || die "could not download ${link}"
+
+    local dest="/usr/local"
```
rafaeldtinoco, line 36: ditto (so you dont have to have /usr/local files, which is a bad practice).

builder/prepare-ubuntu.sh
```diff
@@ -0,0 +1,81 @@
...
+# startup
+
+info "Starting preparation"
+
+install_go
+
+install_pkgs \
+    coreutils bsdutils findutils \
+    build-essential pkgconf \
+    llvm-12 clang-12 \
```
rafaeldtinoco, line 69: See ? Here you're using llvm-XX from the OS (and it is the correct thing since Ubuntu provides many toolchains in its default repository, I would do the same with go).

rafaeldtinoco (commented):
I wouldn't like the "vagrant" experience to be the default (but a complement only), which I think the separate Makefile does.

For me, the default should be a linux box compiling Tracee (and all vagrant recipes in here and in Tracee should be an addendum). 

I would rename builder/prepare-ubuntu.sh to having/specifying it is vagrant related (for other future files not to be confused).

geyslan:
> I wouldn't like the "vagrant" experience to be the default (but a complement only), which I think the separate Makefile does.
> 
> For me, the default should be a linux box compiling Tracee (and all vagrant recipes in here and in Tracee should be an addendum).

@rafaeldtinoco but wouldn't we mixing things this way? I strongly believe that libbpfgo, in despite of having foundations on Tracee, is now a standalone being. So for sure Tracee could make use of a vagrant environment, but its own.

> I would rename builder/prepare-ubuntu.sh to having/specifying it is vagrant related (for other future files not to be confused).

builder/prepare-ubuntu.sh is a loose shell script. It can be reused from a Dockerfile, github workflow etc.

builder/prepare-ubuntu.sh
geyslan: At start I thought having it as an option since we have problems with ubuntu not providing specific tooling versions. But for this case, I assume you're right. I'll check the 22.04 package lists and change it accordingly.

rafaeldtinoco:
> @rafaeldtinoco but wouldn't we mixing things this way? I strongly believe that libbpfgo, in despite of having foundations on Tracee, is now a standalone being. So for sure Tracee could make use of a vagrant environment, but its own.

Sorry, I meant libbpfgo (not Tracee) in my statement.

geyslan:
> Sorry, I meant libbpfgo (not Tracee) in my statement.

@rafaeldtinoco Oh, I get it.

> I wouldn't like the "vagrant" experience to be the default (but a complement only), which I think the separate Makefile does.

So, do you mean create a separate Makefile (e.g. ./contrib/{Makefile,Vagrantfile}) just for controlling the Vagrant box?

grantseltzer (approved):
This looks great to me, nice work!

Makefile
```diff
@@ -180,6 +182,26 @@ helpers-test-static-run: libbpfgo-static
...
+.PHONY: vagrant-destroy
+.PHONY: vagrant-halt
+.PHONY: vagrant-ssh
+
+vagrant-up: .vagrant-up
+vagrant-destroy: .vagrant-destroy
+vagrant-halt: .vagrant-halt
+vagrant-ssh: .vagrant-ssh
+
+.vagrant-%:
```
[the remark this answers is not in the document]
geyslan: Just for letting the reader know that's internal. But they are also not considered first targets: https://www.gnu.org/software/make/manual/make.html#How-Make-Works

Issue #238, opened 2022-09-08
Makefile rule $(VMLINUXH) bpftool err


geyslan:
When bpftool btf dump errs, it may create an empty/malformed $(VMLINUXH) file, making further calls to this rule to wrongly consider it as done.

This issue was discovered via https://github.com/aquasecurity/tracee/pull/2146#issuecomment-1239498911.

Issue #177, opened 2022-06-20
Rewrite and simplify makefiles


grantseltzer:
The makefile system for the selftests and dynamic building is somewhat complex, it can be simplified if we can mirror the makefiles in tracee: https://github.com/aquasecurity/tracee/tree/main/builder

rafaeldtinoco:
Related issues:
https://github.com/aquasecurity/libbpfgo/issues/240
https://github.com/aquasecurity/libbpfgo/issues/242
https://github.com/aquasecurity/libbpfgo/issues/1

PR #239, opened 2022-09-08
makefile: fix vmlinuxh rule

geyslan:
Fixes: #238

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,14 +95,17 @@ $(VMLINUXH): $(OUTPUT)
 		echo "ERROR: kernel does not seem to support BTF"; \
 		exit 1; \
 	fi
-	@if [ ! -f $(VMLINUXH) ]; then \
-		if [ ! $(BPFTOOL) ]; then \
-			echo "ERROR: could not find bpftool"; \
-			exit 1; \
-		fi; \
-		echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
-		$(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH); \
-	fi
+	@if [ ! $(BPFTOOL) ]; then \
+		echo "ERROR: could not find bpftool"; \
+		exit 1; \
+	fi;
+
+	@echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
+	if ! $(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH); then \
+		echo "ERROR: could not create $(VMLINUXH)"; \
+		rm -f "$(VMLINUXH)"; \
+		exit 1; \
+	fi;
 
 # static libbpf generation for the git submodule
 


Makefile
```diff
@@ -95,14 +95,13 @@ $(VMLINUXH): $(OUTPUT)
...
-		echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
-		$(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH); \
-	fi
+	@if [ ! $(BPFTOOL) ]; then \
+		echo "ERROR: could not find bpftool"; \
+		exit 1; \
+	fi;
+
+	@echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
+		$(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH) || rm -f $(VMLINUXH) && exit 1
```
grantseltzer, line 104:
```suggestion
	@echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
	$(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH) || rm -f $(VMLINUXH) && exit 1
```

Can we also add a log saying failed to create vmlinux.h file before exiting?

rafaeldtinoco: BTW, instead of obligating the end user to have bpftool installed (which isn't needed in case of tracee, for example), we could include a "recent enough" vmlinux.h file and rely in libbpf CO-RE functionality to do the relocations.

geyslan: > Can we also add a log saying failed to create vmlinux.h file before exiting?

Ok.

> BTW, instead of obligating the end user to have bpftool installed (which isn't needed in case of tracee, for example), we could include a "recent enough" vmlinux.h file and rely in libbpf CO-RE functionality to do the relocations.

It might work. We need to do some tests. BTW, this PR is a temporary workaround once we'll focus on https://github.com/aquasecurity/libbpfgo/issues/177 later.

Makefile
```diff
@@ -95,14 +95,17 @@ $(VMLINUXH): $(OUTPUT)
...
-			echo "ERROR: could not find bpftool"; \
-			exit 1; \
-		fi; \
-		echo "INFO: generating $(VMLINUXH) from $(BTFFILE)"; \
-		$(BPFTOOL) btf dump file $(BTFFILE) format c > $(VMLINUXH); \
-	fi
+	@if [ ! $(BPFTOOL) ]; then \
+		echo "ERROR: could not find bpftool"; \
+		exit 1; \
+	fi;
```
geyslan, line 101: Despite what was suggested in https://github.com/aquasecurity/libbpfgo/pull/239#discussion_r966337865, this remains for now. Whether we keep an up-to-date vmlinux.h in the tree or build (and make available for consumer) the latest bpftool can be discussed in #177.

My thoughts anyway: the consumer of this library needs, for better experience, an updated tooling (bpftool in this case), which is not offered in common distros, but those that are rolling release.

rafaeldtinoco: Question: why to "fix" this instead of simply adding a vmlinux.h file to the repository (from a recent enough kernel, like 5.15 or so. We can have the file "tar.bz2" and uncompressed by Makefile to save space even. @grantseltzer WDYT ? Obligating the user to have bpftool is complicated. If we pick Ubuntu distribution, for example, it depends on having package named "linux-tools-XXXX" (depending on the kernel version), and is error-prone (sometimes bpftool in a container has wrong version due to host kernel being different, and things like that).

geyslan: Agreeing with you both. I think it demands its own issue, but first let's decides the requirements.

- Remove libbpfgo Makefile bpftool requirement.
- Make available an updated pre-made vmlinux.h (I think it can be one extracted from Arch/Manjaro - due to its freshness) in the libbpfgo tree for static and dynamic compilations (format to be defined).
- Add a Makefile flag (env var) to one specify its own vmlinux.h overriding the offered by libbpfgo.

WDYT?

grantseltzer: @geyslan That sounds great to me!

geyslan: As https://github.com/aquasecurity/libbpfgo/issues/242 was created, I think this PR is ready for review.

geyslan: Done: https://github.com/aquasecurity/libbpfgo/issues/242.

grantseltzer (approved):
LGTM!

PR #241, opened 2022-09-12
kernel_symbols: cache symbols by addr

NDStrahilevitz:
In tracee, kernel symbols need to be fetched by address in potential hot paths.
This change allows this paths to be more optimized.

# Performance analysis

I have run a benchmark of tracee with the following workload
`while true; do strace ps -ef; done`
Running on 7 different cores, with tracee on the 8th.

This is the event distribution of the run:
```
security_socket_connect: 12673 - 0.050712%
hooked_proc_fops: 12835 - 0.051360%
security_bprm_check: 31796 - 0.127235%
task_rename: 31801 - 0.127255%
sched_process_fork: 69034 - 0.276246%
sched_process_exit: 49405 - 0.197699%
container_remove: 3 - 0.000012%
init_namespaces: 1 - 0.000004%
security_file_open: 850949 - 3.405150%
shared_object_loaded: 226939 - 0.908117%
security_socket_create: 12558 - 0.050252%
ptrace: 23624361 - 94.535039%
existing_container: 1 - 0.000004%
container_create: 7 - 0.000028%
sched_process_exec: 31798 - 0.127243%
execve: 35897 - 0.143645%
overall: 24990058
```
The following pprof was the result:
image

As can be seen the function GetSymbolByAddr can be a bottleneck for certain workloads. In tracee's case we run this function for each `hooked_proc_fops` events (see [here](https://github.com/aquasecurity/tracee/blob/main/pkg/ebpf/events_processor.go#L388)). 
Notice - this event is only 0.05% of the workload and results in a 20% cpu time bottleneck.

### DIFF
--- a/helpers/kernel_symbols.go
+++ b/helpers/kernel_symbols.go
@@ -20,8 +20,9 @@ import (
  */
 
 type KernelSymbolTable struct {
-	symbolMap   map[string]KernelSymbol
-	initialized bool
+	symbolMap     map[string]KernelSymbol
+	symbolAddrMap map[uint64]KernelSymbol
+	initialized   bool
 }
 
 type KernelSymbol struct {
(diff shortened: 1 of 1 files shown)


grantseltzer (approved):
LGTM!

PR #253, opened 2022-09-28
c: stdarg: move references to commit, add a note

geyslan:
This clears the original comments inserted by the fix 7c90aa953, moving references to the commit and adding a single note about the required matching calls for va_copy().

From https://man7.org/linux/man-pages/man3/va_arg.3.html:

- _Each invocation of va_start() must be matched by a corresponding invocation of va_end() in the same function._
- _Each invocation of va_copy() must be matched by a corresponding invocation of va_end() in the same function._

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -21,13 +21,15 @@ int libbpf_print_fn(enum libbpf_print_level level,
 	if (level != LIBBPF_WARN)
 		return 0;
 
+	// NOTE: va_list args is managed in libbpf caller
+	// however, these copies must be matched with va_end() in this function
 	va_list exclusivity_check, cgroup_check;
 	va_copy(exclusivity_check, args);
 	va_copy(cgroup_check, args);
-	char *str = va_arg(exclusivity_check, char *);
 
 	// BUG: https://github.com/aquasecurity/tracee/issues/1676
 
+	char *str = va_arg(exclusivity_check, char *);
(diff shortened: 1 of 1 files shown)


javierhonduco:
I would suggest to keep the comments as I think it's unlikely that the next person that modifies this code will check out the past commit messages and introducing issues here is very easy 😞

rafaeldtinoco:
> I would suggest to keep the comments as I think it's unlikely that the next person that modifies this code will check out the past commit messages and introducing issues here is very easy disappointed

IMO commits explaining interfaces used shouldn't be in the code (which is the case). ALSO, a z/OS url (likely related to z/OS USS subsystem) isn't something we want as a reference I suppose.

Either way, we can add back more comments if ever needed when this source is split from the libbpfgo.go file (and hopefully formatted/parsed during checks).

rafaeldtinoco (approved):
LGTM

geyslan:
@javierhonduco I got your point. variadic functions in C are idiosyncratic, however we always need to check documentation along the road. If the comments are part of a well defined doc, it's kinda redundant to be in the code.

I appreciate your concerns. Perhaps to add a single comment to beware the variadic functions as @rafaeldtinoco suggested.

rafaeldtinoco (approved):
LGTM but squash the 2 commits in 1 (IMO)

PR #256, opened 2022-10-05
Save the value of errno before calling printf in cgo code

### LINKED ISSUE #245: Save the value of errno before calling printf in cgo code
In https://github.com/aquasecurity/libbpfgo/pull/244 we discussed how the value of errno should be saved before passing it to printf as printf can change the value.

mozillazg:
Closes #245


geyslan:
LGTM, thanks @mozillazg.

geyslan (approved):

PR #257, opened 2022-10-05
Move embedded C code to a header file

### LINKED ISSUE #250: Move embedded C code to a header file
Since [embedded C code](https://github.com/aquasecurity/libbpfgo/blob/main/libbpfgo.go#L6-L155) is placed as a comment in Go files, we could move into a C header file, enabling the use of tools for automatic formatting and checking following the [Tracee](https://github.com/aquasecurity/tracee/blob/main/builder/Makefile.checkers) way.

This is related to main changes on building - https://github.com/aquasecurity/libbpfgo/issues/177.

However, for this, the goals are only:

- [ ] move the C code to a header to be included in the embedded part.
- [ ] format the new header file using [clang-format](https://github.com/aquasecurity/tracee/blob/main/.clang-format).

mozillazg:
closes #250


grantseltzer:
Thanks for this! Can you run `clang-format -i ./libbpfgo.h` to format the header as well? We need to redo our makefiles but for now can you also add a target for doing so?

mozillazg:
> Can you run clang-format -i ./libbpfgo.h to format the header as well? We need to redo our makefiles but for now can you also add a target for doing so?

@grantseltzer done.

grantseltzer (approved):

PR #258, opened 2022-10-06
Add `InitGlobalVariable` to support setting global variables in bpf code

### LINKED ISSUE #27: eBPF CO-RE improvements: Add support for setting global variables in bpf code
libbpf has a feature for setting global variables in bpf code from userspace. This can be done by accessing a map called ".rodata" and setting values there. The setting happens at load time instead of runtime, making it more performant. Users could rely on this instead of maps or compile-time macros in their bpf projects. 

Relevant mailing list thread: https://lore.kernel.org/bpf/CAO658oVyB2b+Y6K3--sAhTcXfmPpmPjLhA0z7bbjyjhzDV8kcA@mail.gmail.com/

mozillazg:
closes #27


selftest/global-variable/main.go
```diff
@@ -0,0 +1,83 @@
...
+	bpf "github.com/aquasecurity/libbpfgo"
+)
+
+func exitWithErr(err error) {
+	fmt.Fprintln(os.Stderr, err)
+	os.Exit(-1)
+}
+
+func initGlobalVariable(bpfModule *bpf.Module) {
+	rodata, err := bpfModule.GetMap(".rodata")
```
rafaeldtinoco, line 28: Is there a way you think we can support multiple rodata (for multiple variables) like described at: https://lore.kernel.org/bpf/20211021014404.2635234-8-andrii@kernel.org/ ?

mozillazg: I have updated the selftest to cover multiple rodata.

libbpfgo.go
```diff
@@ -776,6 +776,15 @@ func (b *BPFMap) GetValueReadInto(key unsafe.Pointer, value *[]byte) error {
 	return nil
 }
 
+func (b *BPFMap) SetInitialValue(value unsafe.Pointer) error {
+	sz := b.ValueSize()
+	ret := C.bpf_map__set_initial_value(b.bpfMap, value, C.ulong(sz))
```
grantseltzer, line 781: Is this the intended use case of this function? How is using this different from `bpf_map__update_elem`? It would be good contribute the docs to `libbpf.readthedocs.org` for it.

libbpfgo.go
```diff
@@ -776,6 +776,15 @@ func (b *BPFMap) GetValueReadInto(key unsafe.Pointer, value *[]byte) error {
 	return nil
 }
 
+func (b *BPFMap) SetInitialValue(value unsafe.Pointer) error {
```
grantseltzer, line 779: Is there a usecase for this outside of setting global variables? If not maybe we can do something like wrap the `initGlobalVariable()` function you have in the selftest into an API function.

libbpfgo.go
mozillazg: > Is this the intended use case of this function?

Yes, this function can be used to update internal maps: https://lore.kernel.org/bpf/87fszrkhi7.fsf@toke.dk/

> How is using this different from `bpf_map__update_elem`

* `bpf_map__update_elem` update map via `bpf` syscall and only support public maps.
* `bpf_map__set_initial_value` update map via `memcpy` and only support internal maps.

libbpfgo.go
mozillazg: > Is there a usecase for this outside of setting global variables?

No, `bpf_map__set_initial_value` only support this usecase. 

> If not maybe we can do something like wrap the initGlobalVariable() function you have in the selftest into an API function.

I'll remove `SetInitialValue` and add a new API function: 

```go
func (m *Module) InitGlobalVariable(sectionName string, value unsafe.Pointer) error
```

grantseltzer:
I think this approach isn't quite correct. When I've used global variables like this in pure C libbpf, the variables are stored as key/value pairs in a map just called `.rodata`. They way you show your approach working in the selftest creates a new map for each variable with one entry in each map. Screenshot below is testing with a libbpf C native approach:

image

mozillazg:
> I think this approach isn't quite correct. When I've used global variables like this in pure C libbpf, the variables are stored as key/value pairs in a map just called .rodata. The way you show your approach working in the selftest creates a new map for each variable with one entry in each map.

@grantseltzer  We can't init multiple global variables like the pure C way with `skel->rodata->foobar` due to can't update the internet map with a global variable name, so I use multiple internal maps to workaround. Do you know any better way to support multiple global variables?

grantseltzer:
Is it possible to get a handle on the '.rodata' map and iterate over the
keys or just set the values directly with this set initial value API
function?

grantseltzer:
@mozillazg I've opened https://github.com/mozillazg/libbpfgo/pull/2 with a POC of what I have in mind. It works the selftest is, but it fails if you change one of the data types to not match the others (bpf_map__set_initial_value  returns errno 22 (invalid arg)) 

Any thoughts?

mozillazg:
> Is it possible to get a handle on the '.rodata' map and iterate over the
keys 

We can't iterate the map before load bpf object. The '.rodata' map only has one element with 0 as key and the value don't have a well-known struct

> I've opened https://github.com/mozillazg/libbpfgo/pull/2 with a POC of what I have in mind. It works with the selftest as is, but it fails if you change one of the data types to not match the others (bpf_map__set_initial_value returns errno 22 (invalid arg)). Any thoughts?

It looks like we can't use normal way to update the map value. Is it possible to update the memory of value?

```
# parse elf info
elf_info = parse_object("main.bpf.o")

# get global variable info
offset, size = get_global_variable_info(elf_info, "foobar")    # how to do it ??

# get address of map value via bpf_map__initial_value
map_value = bpf_map__initial_value(...)

# update value of global variable via update memory 
update_global_variable(&map_value, offset, &new_foo_value, size)  # how to do it ??
```

grantseltzer:
The error in my approach appears to be an issue with struct memory alignment. If the types add up to a perfect alignment of 8 bytes it works, otherwise it gives the "invalid argument" error.

mozillazg:
> The error in my approach appears to be an issue with struct memory alignment. If the types add up to a perfect alignment of 8 bytes it works, otherwise it gives the "invalid argument" error.

@grantseltzer  👌 

> It looks like we can't use normal way to update the map value. Is it possible to update the memory of value?
> 
> ```
> # parse elf info
> elf_info = parse_object("main.bpf.o")
> 
> # get global variable info
> offset, size = get_global_variable_info(elf_info, "foobar")    # how ??
> 
> # get address of map value via bpf_map__initial_value
> map_value_pointer = bpf_map__initial_value(...)
> 
> # update value of global variable via update memory 
> update_global_variable(map_value_pointer, offset, &new_foo_value, size)  # how ??
> 
> # maybe need to put it back?
> bpf_map__set_initial_value(...)
> ```

A POC of this way: https://github.com/mozillazg/libbpfgo/pull/3

image

mozillazg:
@grantseltzer Which solution do you prefer?

grantseltzer:
@mozillazg  Your solution in https://github.com/mozillazg/libbpfgo/pull/3 works great! Nicely done, really enjoyed collaborating with you on this!

Only thing is I kindly ask you remove the print statements and get the selftest working (remove time.Sleep). Also my suggestion for the GoDoc comment about InitGlobalVariable to something like:

```
// InitGlobalVariable sets global variables (defined in .data or .rodata)
// in bpf code. It must be called before the BPF object is loaded.

```

mozillazg:
@grantseltzer updated codes to use solution like https://github.com/mozillazg/libbpfgo/pull/3 . please review it again.

grantseltzer (approved):
Looks great, well done and thank you very much @mozillazg!

PR #261, opened 2022-10-12
add unit tests for helpers/osinfo.go

### LINKED ISSUE #124: Add unit tests for osinfo.go
https://github.com/aquasecurity/libbpfgo/blob/main/helpers/osinfo.go seems to be untested. It would be helpful to add some unit tests to cover certain scenarios.

ronaudinho:
An attempt to resolve #124 but it seems that most of the functions requires checking for existing path in the OS, not sure how to simulate them. Some tests seem superfluous too, will comment on changes for details.

### DIFF
--- a/helpers/osinfo_test.go
+++ b/helpers/osinfo_test.go
@@ -1,0 +1,164 @@
+package helpers
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOSReleaseID_String(t *testing.T) {
+	testCases := []struct {
+		testName string
+		in       uint32
+		want     string
+	}{
+		{
+			testName: "happy path",
+			in:       1,
+			want:     "ubuntu",
+		},
+		{
+			testName: "no exists",
+			in:       10,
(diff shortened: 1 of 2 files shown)


helpers/osinfo_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOSReleaseID_String(t *testing.T) {
```
ronaudinho, line 11: only tests for the existence of value in map, i don't feel this is necessary. will remove if not needed, else will add `TestOSReleaseField_String` later.

grantseltzer:
Hi @ronaudinho ! Thanks so much for this contribution, i'll review this by end of week!

grantseltzer:
Woops, left that review from the bot account while working on it. That was from me hahaha

rafaeldtinoco:
Resolved some comments, changed minor things and committed with credits at: https://github.com/aquasecurity/libbpfgo/pull/278. Closing this one.

PR #263, opened 2022-10-13
Redefine BPFProgType constants and remove libbpfgo import

grantseltzer:
This completely removes dependence between either helpers or libbpfgo

### DIFF
--- a/helpers/argumentParsers.go
+++ b/helpers/argumentParsers.go
@@ -3,12 +3,11 @@ package helpers
 import (
 	"encoding/binary"
 	"fmt"
-	"golang.org/x/sys/unix"
 	"net"
 	"strconv"
 	"strings"
 
-	"github.com/aquasecurity/libbpfgo"
+	"golang.org/x/sys/unix"
 )
 
 type SystemFunctionArgument interface {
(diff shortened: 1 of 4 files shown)


helpers/argumentParsers.go
```diff
@@ -1708,45 +1707,127 @@ func ParseGetSocketOption(rawValue uint64) (SocketOptionArgument, error) {
...
+	BPFProgTypeSkLookup.Value():              BPFProgTypeSkLookup,
+	BPFProgTypeSyscall.Value():               BPFProgTypeSyscall,
 }
 
-func ParseBPFProgType(rawValue uint64) (libbpfgo.BPFProgType, error) {
+func ParseBPFProgType(rawValue uint64) (BPFProgType, error) {
 	v, ok := bpfProgTypeMap[rawValue]
 	if !ok {
-		return libbpfgo.BPFProgType(0), fmt.Errorf("not a valid argument: %d", rawValue)
+		return BPFProgType(0), fmt.Errorf("not a valid argument: %d", rawValue)
```
geyslan, line 1830: What about instead of another map we implement only one following this struct logic:

```go
const (
	BPFProgTypeUnspec uint64 = iota
	BPFProgTypeSocketFilter
)

type BPFProgType struct {
	name  string
	value uint64
}

var bpfProgTypeMap = map[uint64]BPFProgType{
	BPFProgTypeUnspec: {
		name:  "BPF_PROG_TYPE_UNSPEC",
		value: BPFProgTypeUnspec,
	},
	BPFProgTypeSocketFilter: {
		name: "BPF_PROG_TYPE_SOCKET_FILTER",
		value: BPFProgTypeSocketFilter,
	},
}

func (b BPFProgType) String() string {
	return b.name
}

func (b BPFProgType) Value() uint64 {
	return b.value
}
```

geyslan (approved):
LGTM, I put some comments with nits and suggestions not to be taken as change requests.

helpers/argumentParsers.go
geyslan: Oh, I think I got it. Keeping the Name string is not required.

Said that, why not just to make a conversion here and get rid of the the map that converts from uint64 to uint32?

```go
func ParseBPFProgType(rawValue uint64) (BPFProgType, error) {
	v, ok := bpfProgTypeMap[uint32(rawValue)]
```

grantseltzer: I see what you mean, I believe uint64 was purposefully used as a parsing parameter because thats how the value is read from raw events, not 100% sure though.

PR #265, opened 2022-10-25
fix: re-introduce memory bumping

rafaeldtinoco:
make libbpfgo to bump rlimits instead of relying in libbpf to do so.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -296,6 +296,20 @@ func (b LibbpfStrictMode) String() (str string) {
 	return str
 }
 
+// NOTE: libbpf has started raising limits by default but, unfortunately, that
+// seems to be failing in current libbpf version. The memory limit bump might be
+// removed once this is sorted out.
+func bumpMemlockRlimit() error {
+	var rLimit syscall.Rlimit
+	rLimit.Max = 512 << 20 /* 512 MBs */
+	rLimit.Cur = 512 << 20 /* 512 MBs */
+	err := syscall.Setrlimit(C.RLIMIT_MEMLOCK, &rLimit)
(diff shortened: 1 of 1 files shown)


geyslan (approved):

kakkoyun:
Hey @rafaeldtinoco, what was the reason for this revert? I couldn't find it. 

Would you also be open to making it configurable?  We'd love to disable it or find a way to specify the value of the lock. I'd be happy to contribute. I want to understand the reasoning and possible downsides.

rafaeldtinoco:
> Hey @rafaeldtinoco, what was the reason for this revert? I couldn't find it.
> 
> Would you also be open to making it configurable? We'd love to disable it or find a way to specify the value of the lock. I'd be happy to contribute. I want to understand the reasoning and possible downsides.

@grantseltzer found out that libbpf wouldn't raise the limit in all cases and I believe he reported an issue upstrea, did you @grantseltzer ? Could you remind me what was the issue with libbpf missing the raise in some environments ?

grantseltzer:
@kakkoyun @rafaeldtinoco  Yes I brought it up on the mailing list, you can find the thread [here](https://lore.kernel.org/bpf/CAEf4BzaMNU2UoutOLdX-HbkA0037EV6kwQg3-TJLHV_8tjdzDg@mail.gmail.com/) but essentially the summary is that it's a known issue on certain kernels to not be able to properly detect if the limit should manually be raised.

rafaeldtinoco:
@kakkoyun feel free to suggest something to turn it into a tunable parameter!

bobrik:
I made #337 to skip the bump on a tunable condition.

PR #267, opened 2022-11-06
Add parsing function for mmap flags argument

AlonZivony:


helpers/argumentParsers.go
```diff
@@ -1833,3 +1832,121 @@ func ParseBPFProgType(rawValue uint64) (BPFProgType, error) {
...
+	MapGrowsdown      MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_GROWSDOWN, stringValue: "MAP_GROWSDOWN"}
+	MapDenywrite      MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_DENYWRITE, stringValue: "MAP_DENYWRITE"}
+	MapExecutable     MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_EXECUTABLE, stringValue: "MAP_EXECUTABLE"}
+	MapLocked         MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_LOCKED, stringValue: "MAP_LOCKED"}
+	MapNoreserve      MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_NORESERVE, stringValue: "MAP_NORESERVE"}
+	MapFile           MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_FILE, stringValue: "MAP_FILE"}
+	MapHuge2MB        MmapFlagArgument = MmapFlagArgument{rawValue: 21 << HugetlbFlagEncodeShift, stringValue: "MAP_HUGE_2MB"}
+	MapHuge1GB        MmapFlagArgument = MmapFlagArgument{rawValue: 30 << HugetlbFlagEncodeShift, stringValue: "MAP_HUGE_1GB"}
+	MapSYNC           MmapFlagArgument = MmapFlagArgument{rawValue: unix.MAP_SYNC, stringValue: "MAP_SYNC"}
+	// TODO: Add support for MAP_UNINITIALIZED which collide with Huge TLB size bits
```
grantseltzer, line 1869: Why not implement this now?

AlonZivony: I think there is a kernel bug here - the value of `MAP_UNINITIALIZED` collide with the bits dedicated to the `HUGETLB` size.
So we can't parse them together.
The  `MAP_UNINITIALIZED` flag isn't usable unless you config the kernel to allow it, which is advised strongly against in the documentation because of security reasons.

grantseltzer (approved):
LGTM

PR #268, opened 2022-11-11
Add libbpfgo_full_static tag for 'fully' static builds

jeffmahoney:
The static builds we have now don't require an external libbpf at runtime but do require an external libelf and zlib.  I have a project where we're building BPF modules into an executable to be distributed on client hosts where no additional configuration must be required.  Pretty much every system has zlib but few under regular operation have libelf installed.

This commit adds a new build mode: 'full static' where there are no external runtime dependencies (other than libc).  The user must build their project with the libbpfgo_full_static tag defined and that's it.

The libbpfgo project now requires go 1.18 so features like ${SRC_DIR} in CFLAGS/LDFLAGS eliminate the need for specifying those via environment variables.  There's no reason this approach couldn't be used for dynamic and regular static build but I didn't want to break existing projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,22 @@ libbpfgo-static-test: libbpfgo-test-bpf-static
 		-v -tags netgo -ldflags $(CGO_EXTLDFLAGS_STATIC) \
 		.
 
+libbpfgo-full-static: $(VMLINUXH) | $(LIBBPF_OBJ)
+	CC=$(CLANG) \
+		GOOS=linux GOARCH=$(ARCH) \
+		$(GO) build \
+		-tags 'netgo libbpfgo_full_static' -ldflags $(CGO_EXTLDFLAGS_STATIC) \
+		.
+
+libbpfgo-full-static-test: libbpfgo-test-bpf-static
+	sudo env PATH=$(PATH) \
+		CC=$(CLANG) \
+		GOOS=linux GOARCH=$(ARCH) \
+		$(GO) test \
+		-v -tags 'netgo libbpfgo_full_static' -ldflags $(CGO_EXTLDFLAGS_STATIC) \
+		.
+
+
 # vmlinux header file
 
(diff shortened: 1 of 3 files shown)


grantseltzer:
Hi @jeffmahoney! Thanks so much for contributing this. I previously overlooked this issue, do you think there's a need to have both of these targets (just libbpf static, vs fully static)? My initial thought is that we can just convert the current 'static' build to be fully static.

jeffmahoney:
Sure, I can pull in the static zlib too. Ultimately, we'll end up pulling a dynamic libc anyway, though. So it's a judgement call whether depending on zlib being there is a requirement less frequently met than just having libc. I ended up making the call that we could pretty much depend on zlib being there.

grantseltzer:
Just want to understand, why do we still have to pull in dynamic libc?

jeffmahoney:
As I understand it, using cgo results in a dynamic binary. There are parts of glibc that don't work when built into a static binary, though they might not matter here. I haven't tried forcing my entire project static, though.

yanivagman:
In tracee we use the static flavor of libbpfgo to build an entirely static binary (when chosen).
To do that, we also use the static libs of libelf and zlib.

rafaeldtinoco:
Yep, unfortunately, looks like Fedora does not provide static versions of libelf and zlib anymore (they used to in the past). Some other distributions also don't. Debian and Ubuntu do. In tracee, we can compile with "ubuntu-make" and it compiles tracee statically inside an ubuntu building environment (thus it does not matter if libelf static libs aren't available).

Maybe it is not a bad idea to allow libelf to be included, it is worth the discussion at least.

rafaeldtinoco:
Okay, after some discussions we decided not to include other branches as submodules. This way, it is recommended that, in order to get static builds, a docker building environment (or similar approach) is taken. Thank you!

PR #273, opened 2022-12-12
libbpf: fix access violation in libbpf_print_fn.

rafaeldtinoco:
libbpf_print_fn uses va_arg and explicit argument type "char*" to retrieve arguments as strings. This assumption is incorrect.  In case when libbpf logs a printf format string with argument being a non-string type, the str pointer will point to memory that is not NULL terminated. Wnen strstr is called with a non-NULL terminated string, we hit access violation on the stack.

This change uses a temporary string allocated on the stack to be filled with a printf formatted string with the arguments converted into the string. The keywords search are performed on the temporary string buffer.

Signed-off-by: Hao Xiang <haoxiang@bytedance.com>

### DIFF
--- a/libbpfgo.h
+++ b/libbpfgo.h
@@ -11,40 +11,39 @@
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
 
-int libbpf_print_fn(enum libbpf_print_level level, const char *format,
-                    va_list args) {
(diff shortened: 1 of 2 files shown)


rafaeldtinoco:
Related issue: https://github.com/aquasecurity/tracee/issues/2446

geyslan (dismissed):
LGTM

PR #281, opened 2022-12-22
Fix/kernel symbol memory

NDStrahilevitz:
Some memory optimizations and fixes in the `kernel_symbols.go` helpers.

```
    1. helpers: share kernel symbol between maps
    
    The KernelSymbolTable contains two map caches for each potential search
    one per name and one per address.
    As such the symbols are copied twice for each map.
    Fix this by making the maps values a *KernelSumbol so one object will
    be shared between them.
    
    2. helpers: fix string allocations in NewKernelSymbolsMap
    
    String allocations could dangle due to being assigned substrings.
    Fix it by using strings.Clone.
    Also use simple string concatenation instead of fmt.Sprintf to avoid
    related allocations.
```

### DIFF
--- a/helpers/kernel_symbols.go
+++ b/helpers/kernel_symbols.go
@@ -57,19 +57,19 @@ func NewKernelSymbolsMap() (*KernelSymbolTable, error) {
 		if err != nil {
 			continue
 		}
-		symbolType := line[1]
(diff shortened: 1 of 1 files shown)


yanivagman (approved):
LGTM

PR #282, opened 2022-12-26
Lazy kernel symbols

NDStrahilevitz:
Lazy kernel symbols is a slower but more memory efficient of the
KernelSymbolsTable (which is now an interface).
It works by storing a copy of the kallsyms file and querying it's content,
instead of preparsing the whole file at construction time.

Also fix a small leak in the `SymbolToOffset` helper where a file `Close()` was missing.


NDStrahilevitz:
@yanivagman tests are failing because they need root/syslog caps to run, otherwise reading `kallsyms` yields the results with all addresses set to `0`.

yanivagman (approved):
LGTM

yanivagman (approved):

PR #283, opened 2023-01-05
lazy ksymbols: fix address search

NDStrahilevitz:
ksymbols may sometimes not be sorted at the edges, usually for non system symbols. As such searching by binary search may fail. Add a reverse linear search at the end to catch these symbols.

```
In commit 3ea92310a0, a new KernelSymbolsTable implementation was added
which uses lazy querying and parsing of a stored string copy of the
/proc/kallsyms file.

In order to keep efficiency of non cached queries, binary search was
used for querying symbols by address, since the file mostly stores the
symbols in order of addresses.

However, on kernels compiled with CONFIG_KALLSYMS_ALL=y, symbols from
other sections may be present in a non ordered place. This wasn't
taken in consideration in initial implementations, and so a query by
address could have failed despite existing.
These symbols usually appear near the end of the file, so a reverse
linear search is added after a failed binary search.
```

### DIFF
--- a/helpers/kernel_symbols.go
+++ b/helpers/kernel_symbols.go
@@ -51,6 +51,22 @@ func symbolKey(owner, name string) string {
 	return owner + "_" + name
 }
 
+// parses ksymbol line file file. returns in order the symbols "type", "name", "order"
+func parseSymbolLine(line []string) (string, string, string) {
+	symbolType := strings.Clone(line[1])
+	symbolName := strings.Clone(line[2])
+
+	symbolOwner := "system"
+	if len(line) > 3 {
+		// When a symbol is contained in a kernel module, it will be specified
+		// within square brackets, otherwise it's part of the system
+		symbolOwner = strings.Clone(line[3])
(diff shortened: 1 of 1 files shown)


geyslan (approved):
LGTM.

geyslan:
> What about to sort `fileLines` in the `lazyKernelSymbols.Refresh()` just after the reading? Perhaps we can stick with the binary search only.

In addition to that possibility, I think we can refactor a bit by moving some of the parsing logic into a function.

geyslan (dismissed):

rafaeldtinoco:
Could you explain better in the git log what was the fix you did with this commit? Also, you recommended a PR in tracee to use lazy symbols, if it is broken it is better to update that PR as well (asking for the author to include this fix). Thanks!

rafaeldtinoco (changes requested):
Still need to review, but asked for a better git log.

rafaeldtinoco (approved):
LGTM

Issue #285, opened 2023-01-13
libbpf_print_fn: va_arg mismatched pair


geyslan:
`libbpf_print_fn()` has an mismatched pair of va_copy() and va_end() calls.

PR #286, opened 2023-01-13
fix: va_end() va_list copy rather than copied

geyslan:
> Each invocation of va_copy() must be matched by a corresponding invocation of va_end() in the same function.

https://man7.org/linux/man-pages/man3/va_arg.3.html

Fix:
- #285


yanivagman (approved):
LGTM

PR #287, opened 2023-01-31
fix hardcoded flag values

NDStrahilevitz:
Calls to the `open` syscall used hardcoded flag values. Since flag values may differ between arches this could cause EINVAL results.


yanivagman (approved):
LGTM

aymericDD:
Thanks for this fix 🙏 
Does anyone know when this fix will be released?

Issue #293, opened 2023-02-20
Remove default filter callbacks


geyslan:
## Description

As we now have libbpfgo log callbacks, the logic for filtering must reside in the consumer.

- [ ] remove default filter callbacks
- [ ] change tests to better exemplify the callbacks usage

## Additional details 

Context:

- https://github.com/aquasecurity/libbpfgo/pull/284

PR #294, opened 2023-02-22
logger_cb: remove default callbacks

geyslan:
The logic related to filtering by callbacks must be a decision of the library consumers, as it is not possible to know which is the best strategy for each case.

This also adds to the related selftest other example that shows how to filter using callbacks.

Fixes: #293

- #293

### DIFF
--- a/logger_cb.go
+++ b/logger_cb.go
@@ -8,7 +8,6 @@ import "C"
 import (
 	"fmt"
 	"os"
-	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


logger_cb.go
```diff
@@ -47,102 +46,30 @@ const (
 
 // Callbacks stores the callbacks to be used by libbpfgo
 type Callbacks struct {
-	Log        func(level int, msg string, keyValues ...interface{})
+	Log        func(level int, msg string)
```
geyslan, line 49: We don't need to worry about what's not relevant to libbpf.

rafaeldtinoco (approved):
LGTM

geyslan:
> LGTM

Thanks, merging in minutes.

PR #295, opened 2023-02-28
ci: Use Go 1.20

javierhonduco:
Golang 1.20 was released early this month https://go.dev/blog/go1.20, bringing some nice perf improvements:

> Improved performance
> Compiler and garbage collector improvements have reduced memory overhead and improved overall CPU performance by up to 2%.
> Work specifically targeting compilation times led to build improvements by up to 10%. This brings build speeds back in line with Go 1.17.

### DIFF
--- a/.github/actions/build-dependencies/action.yaml
+++ b/.github/actions/build-dependencies/action.yaml
@@ -7,7 +7,7 @@ runs:
     - name: Setup Go
       uses: actions/setup-go@v2
       with:
-        go-version: "1.18"
+        go-version: "1.20"
     - name: Install Compilers & Formatters
       run: |
         sudo apt-get update


geyslan:
@javierhonduco Thanks for this PR.

I think other changes coming in 1.20 will be a great add like:

> The new [context.WithCancelCause](https://go.dev/pkg/context#WithCancelCause) function provides a way to cancel a context with a given error. That error can be retrieved by calling the new [context.Cause](https://go.dev/pkg/context#Cause) function.

@rafaeldtinoco do we have any current golang version restrictions in tracee?

rafaeldtinoco:
We're still on go1.19 there. IMO, the correct thing would be for us to test a few golang versions (not a single one) with the selftests. This way the library consumers would know libbpfgo works with go1.17, go1.18, 1.19 and 1.20, for example.

javierhonduco:
> IMO, the correct thing would be for us to test a few golang versions (not a single one)

Makes sense, happy to close this PR then

rafaeldtinoco:
@javierhonduco opened https://github.com/aquasecurity/libbpfgo/issues/296 for the need. Thanks a lot!

Issue #300, opened 2023-03-15
Large libbpf messages are truncated


javierhonduco:
Hi,

Recently we've seen that long verifier messages that used to be displayed in their entirety are now truncated. This unfortunately hinders debugging verifier problems. An example in https://github.com/javierhonduco/parca-agent/tree/truncated-logs-repro, built and ran with:

``` 
$ make && make debug/build && sudo dist/parca-agent-debug --node=test --log-level="debug"  --verbose-bpf-logging
```

fails with these truncated logs

``` 
libbpf: prog 'walk_user_stacktrace_impl': BPF program load failed: Invalid argument
libbpf: prog 'walk_user_stacktrace_impl': -- BEGIN PROG LOAD LOG --
0: R1=ctx(off=0,imm=0) R10=fp0
; int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) {
0: (bf) r9 = r1                       ; R1=ctx(off=0,imm=0) R9_w=ctx(off=0,imm=0)
; u64 pid_tgid = bpf_get_current_pid_tgid();
1: (85
libbpf: prog 'walk_user_stacktrace_impl': failed to load: -22
libbpf: failed to load object 'parca'
```

I believe this behaviour was introduced in https://github.com/aquasecurity/libbpfgo/commit/e3f69db9956650aa3e5d08624965915ba750086d. The buffer is now 300 chars (including nul).

>   out = (char *)calloc(1, 300);

```python
>>> len("libbpf: prog 'walk_user_stacktrace_impl': -- BEGIN PROG LOAD LOG -- 0: R1=ctx(off=0,imm=0) R10=fp0 ; int walk_user_stacktrace_impl(struct bpf_perf_event_data *ctx) { 0: (bf) r9 = r1                       ; R1=ctx(off=0,imm=0) R9_w=ctx(off=0,imm=0) ; u64 pid_tgid = bpf_get_current_pid_tgid(); 1: (85")
299
```

cc/ @kakkoyun @Sylfrena @brancz

geyslan:
@javierhonduco thanks for bringing that. Can you please test increasing that buffer from your end?

javierhonduco:
With a buffer size of 1000:

```patch
-  out = (char *)calloc(1, 300);
+  out = (char *)calloc(1, 1000);
  if (!out)
    return -ENOMEM;

  va_copy(check, args);
-  ret = vsnprintf(out, 300, format, check);
[...]

PR #301, opened 2023-03-19
fix truncate output

geyslan:
We were incorrectly using a fixed-length buffer to hold the output. This allocates a sufficiently large buffer.

Fix: #300

### DIFF
--- a/libbpfgo.h
+++ b/libbpfgo.h
@@ -22,16 +22,25 @@ int libbpf_print_fn(enum libbpf_print_level level, // libbpf print level
                     const char *format,            // format used for the msg
                     va_list args) {                // args used by format
 
-  int ret = 0;
+  int ret;
+  size_t len;
   char *out;
   va_list check;
 
-  out = (char *)calloc(1, 300);
+  va_copy(check, args);
+  ret = vsnprintf(NULL, 0, format, check); // get output length
+  va_end(check);
+
+  if (ret < 0)
+    return ret;
+
+  len = ret + 1; // add 1 for NUL
+  out = malloc(len);
   if (!out)
     return -ENOMEM;
 
   va_copy(check, args);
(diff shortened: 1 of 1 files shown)


libbpfgo.h
```diff
@@ -22,16 +22,25 @@ int libbpf_print_fn(enum libbpf_print_level level, // libbpf print level
...
-  int ret = 0;
+  int ret;
+  size_t len;
   char *out;
   va_list check;
 
-  out = (char *)calloc(1, 300);
+  va_copy(check, args);
+  ret = vsnprintf(NULL, 0, format, check); // get output length
+  va_end(check);
```
yanivagman, line 32: I didn't use va_args much in the past, but do we need to call va_copy and va_end twice in this function (here and below when using vsnprintf)?

geyslan: stdarg man stands:

```
...
If ap is passed to a function that uses va_arg(ap,type), then the value of ap is undefined after the return of that function.
```

So, being cautious, as the first call of vsnprintf iterates all args, I assume it's best to reset it.

geyslan: vsnprintf man:

```
The functions vprintf(), vfprintf(), vdprintf(), vsprintf(), vsnprintf() are equivalent to the functions printf(), fprintf(), dprintf(), sprintf(), snprintf(), respectively, except that they are called with a va_list instead of a variable number of arguments. These functions do not call the va_end macro. Because they invoke the va_arg macro, the value of ap is undefined after the call. See stdarg(3).
```

yanivagman: What I thought is that maybe you can call va_copy once (where you call it first now), and only leave the last va_end.
But sice I didn't read the man of va_args and I believe you did, you can ignore this comment if it is not relevant.

yanivagman (approved):
LGTM

PR #254, opened 2022-10-04
Add `AttachIter` to wrap `bpf_program__attach_iter`

mozillazg:
https://github.com/aquasecurity/libbpfgo/issues/91

PR #304, opened 2023-03-23
chore: call bpf_program__attach_iter from go

geyslan:
This changes libbpfgo to call `bpf_program__attach_iter` directly from go instead of doing this via a C wrapper. What this means is that we maintained opts creation in C, but the actual libbpf attach call is done in go.

This creates two new functions in libbpfgo.h as helpers:
- `bpf_iter_attach_opts_new()`
- `bpf_iter_attach_opts_free()`


geyslan:
As questioned in https://github.com/aquasecurity/libbpfgo/pull/254#discussion_r1144230843.

rafaeldtinoco (approved):
Thanks for that @geyslan. My OCD sent compliments!

geyslan:
> Thanks for that @geyslan. My OCD sent compliments!

Issue #305, opened 2023-03-23
bump libbpf to map creation upstream fix


geyslan:
[libbpf v1.1.0](https://github.com/libbpf/libbpf/releases/tag/v1.1.0) is failing with kernels without `BPF_F_MMAPABLE` support.

libbpf must be bumped to the upstream:
- ~https://github.com/libbpf/libbpf/commit/ce8d078ac762df56a8407456601debd25457672e~ -> https://github.com/libbpf/libbpf/commit/b2e29a1 (see comment below)

Context:
- https://github.com/aquasecurity/tracee/pull/2911#issuecomment-1481473694

geyslan:
We can't checkout https://github.com/libbpf/libbpf/commit/ce8d078ac762df56a8407456601debd25457672e, since there were other changes related to map what bumped it to 1.2 while libbpf remained 1.1. See [build](https://github.com/aquasecurity/libbpfgo/actions/runs/4506107048/jobs/7932538316?pr=306) error:

![image](https://user-images.githubusercontent.com/3372117/227386416-5e3b34ae-e420-4450-a4c4-f2789c36c246.png)

So the actual bump needs to be to 1.2 - https://github.com/libbpf/libbpf/commit/b2e29a1.

PR #306, opened 2023-03-23
libbpf: bump to b2e29a1

geyslan:
This is a minor bump to include the following libbpf fix (and all
the related changes in libbpf submodule):

```
  libbpf: Fix map creation flags sanitization

  As BPF_F_MMAPABLE flag is now conditionnaly set (by map_is_mmapable),
  it should not be toggled but disabled if not supported by kernel.
```

Fixes: #305


rafaeldtinoco (approved):
LGTM

PR #308, opened 2023-03-26
clang-format: add .clang-format file

geyslan:
Add a .clang-format file to the root of the repository. This file is used by clang-format to format the code. The file is based on the Tracee style, with some modifications to make it more suitable for libbpfgo.

This commit also formats the code using the new .clang-format file.


yanivagman (approved):
LGTM

PR #309, opened 2023-03-27
Add explicit timeout to perf + ring buffer polling

javierhonduco:
Right now it's hardcoded to 300ms which might not work for every use-case. Having it as a parameter allows users to decide their tradeoff between latency and overhead.

cc @kakkoyun


yanivagman (approved):
Totally makes sense to add user-defined timeout.
Thanks @javierhonduco LGTM!

rafaeldtinoco (approved):
LGTM

Issue #311, opened 2023-03-29
log callback isn't passing the right output level


geyslan:
When dealing with https://github.com/aquasecurity/tracee/pull/2933, I found that libbpfgo is passing, via callback, an output level that is always zero.

https://github.com/aquasecurity/libbpfgo/blob/8c4c96dfe6ace9c8ea3eef4333d9f1c74fb32db0/logger_cb.go#L37

PR #312, opened 2023-03-29
logger: fix callback log level and output

geyslan:
commit e99ef7235774fcbdfa3d3f769a18f9fbbbf3bf69

    logger: pass raw libbpf print output
    
    Due to libbpf output being escaped, we should pass the raw output to the
    consumer, letting them decide how to format it, since them might want
    to, for example, use it in a logger or print directly to stderr.
    
    Reported-by: @yanivagman

commit 9f21a9d91dc47c3152c95c7899fe7493e0d23452

    logger: fix callback log level

Fixes: #311
Fixes: #313

### DIFF
--- a/logger_cb.go
+++ b/logger_cb.go
@@ -8,7 +8,6 @@ import "C"
 import (
 	"fmt"
 	"os"
-	"strings"
 )
 
 // This callback definition needs to be in a different file from where it is declared in C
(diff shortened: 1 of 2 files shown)


yanivagman (dismissed):
LGTM

geyslan:
> LGTM

There's a new commit addressing also https://github.com/aquasecurity/libbpfgo/issues/313. Would be great to have local tests confirmation before merging it.

geyslan:
@yanivagman I guess we can't output it as literal strings (to be used by a json structured logger): https://github.com/uber-go/zap/issues/963#issuecomment-859978457

PR #314, opened 2023-04-03
chore: call bpf_object__open_mem from go

geyslan:
This changes libbpfgo to call `bpf_object__open_mem` directly from go instead of doing this via a C wrapper. What this means is that we maintained opts creation in C, but the actual libbpf object open call is done in go.

This creates two new functions in libbpfgo.h as helpers:
- `bpf_object_open_opts_new()`
- `bpf_object_open_opts_free()`

Context: #304

Additionally, this commit enforces the use of defer and resolves a potential `CKconfigPath` leak when `len(args.KConfigFilePath)` is > 2.


rafaeldtinoco:
rebased at: https://github.com/aquasecurity/libbpfgo/pull/325 (keeping credits).

Issue #195, opened 2022-07-24
libbpf_num_possible_cpus support ?


rafaeldtinoco:
Maybe we should add support to:

![image](https://user-images.githubusercontent.com/7395852/180629121-c9913340-3c72-4938-a7c9-776b1288c04e.png)

and improve the experience when trying to get which CPUs are online..

Discussion at: https://github.com/aquasecurity/libbpfgo/issues/191

rafaeldtinoco:
All libbpf/libbpfgo users relying in per CPU map feature will have to do some logic like this:

![image](https://user-images.githubusercontent.com/7395852/180629182-ac462f93-5e55-4896-ad91-3941db76cda5.png)

javierhonduco:
Adding this API would be great! Right now some projects rely on Go's [runtime library](https://pkg.go.dev/runtime#NumCPU), but I am not sure if that accounts for offline CPUs as libbpf does

rafaeldtinoco:
> Adding this API would be great! Right now some projects rely on Go's [runtime library](https://pkg.go.dev/runtime#NumCPU), but I am not sure if that accounts for offline CPUs as libbpf does

No it does not I think. Knowing how many CPUs are enabled or not is not enough for when dealing with per-cpu maps. We need to know exact CPU number that is online to know from which buffer to pick the info afaict.

javierhonduco:
Makes sense, will move to `libbpf_num_possible_cpus` as soon as it's available 😄

v-thakkar:
If nobody is working on this, I'll be happy to tackle this. It'll be nice to use libbpf_num_possible_cpus instead of runtime.NumCPU() in parca-agent especially because it doesn't seem to be reflecting the changes to the CPU allocation after the process has started.

PR #331, opened 2023-05-31
Expose libbpf_num_possible_cpus

javierhonduco:
To fetch the number of "possible" (`/sys/devices/system/cpu/possible`) CPUs.

This is needed in various PERCPU structures to compute the value size. More details in https://github.com/parca-dev/parca-agent/issues/1696#issuecomment-1568736923

Fixes: https://github.com/aquasecurity/libbpfgo/issues/195

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -2089,3 +2089,11 @@ func BPFProgramTypeIsSupported(progType BPFProgType) (bool, error) {
 	}
 	return cSupported == 1, nil
 }
+
+func NumPossibleCPUs() (int, error) {
+	numCPUs, _ := C.libbpf_num_possible_cpus()
+	if numCPUs < 0 {
+		return 0, fmt.Errorf("failed to retrieve the number of CPUs")
+	}
+	return int(numCPUs), nil
+}


rafaeldtinoco (approved):
LGTM

libbpfgo.go
```diff
@@ -2089,3 +2089,11 @@ func BPFProgramTypeIsSupported(progType BPFProgType) (bool, error) {
 	}
 	return cSupported == 1, nil
 }
+
+func NumPossibleCPUs() (int, error) {
+	numCPUs, _ := C.libbpf_num_possible_cpus()
+	if numCPUs < 0 {
+		return 0, fmt.Errorf("failed to retrieve the number of CPUs")
```
geyslan, line 2096: Sorry for the late review.

Shouldn't this error have errno information? For example:

https://github.com/aquasecurity/libbpfgo/blob/7c15abd453664293a200801392bfcc87162fe5b4/libbpfgo.go#L928-L930C3

geyslan: No worries at all.

geyslan: > I don't think we need to check for errno here, the only error path from the function is [here](https://github.com/libbpf/libbpf/blob/f7eb43b90f4c8882edf6354f8585094f8f3aade0/src/libbpf.c#LL12631C3-L12631C9), but let me know if you think that's not correct

We standardized returns taking into account possible changes in libbpf. Perhaps currently libbpf does not have a clear set of errno in this function, but it may be different in the future. And above all, as cgo provides us errno in C calls is always good to make use of it.

geyslan:
@javierhonduco thanks for your contribution. It was shipped in [v0.4.9-libbpf-1.2.0](https://github.com/aquasecurity/libbpfgo/releases/tag/v0.4.9-libbpf-1.2.0).

PR #332, opened 2023-06-02
error: consider errno as a return value

geyslan:
CGo provide us a way to get errno value from C code as a return value, so it's always a good idea to make use of it.

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -2091,9 +2091,9 @@ func BPFProgramTypeIsSupported(progType BPFProgType) (bool, error) {
 }
 
 func NumPossibleCPUs() (int, error) {
-	numCPUs, _ := C.libbpf_num_possible_cpus()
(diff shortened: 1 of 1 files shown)


libbpfgo.go
```diff
@@ -2091,9 +2091,9 @@ func BPFProgramTypeIsSupported(progType BPFProgType) (bool, error) {
 }
 
 func NumPossibleCPUs() (int, error) {
-	numCPUs, _ := C.libbpf_num_possible_cpus()
+	numCPUs, errC := C.libbpf_num_possible_cpus()
 	if numCPUs < 0 {
-		return 0, fmt.Errorf("failed to retrieve the number of CPUs")
+		return 0, fmt.Errorf("failed to retrieve the number of CPUs: %w", errC)
```
javierhonduco, line 2096: Was just gonna do this, you were faster than me 😄. Thanks!

Issue #91, opened 2021-11-12
Implement libbpf bpf_program API


grantseltzer:
- [ ] bpf_program__prev
- [ ] bpf_program__set_priv
- [ ] bpf_program__priv
- [ ] bpf_program__set_ifindex
- [ ] bpf_program__name
- [ ] bpf_program__section_name
- [ ] bpf_program__autoload
- [ ] bpf_program__set_autoload
- [ ] bpf_program__fd
- [ ] bpf_program__pin
- [ ] bpf_program__unpin
- [ ] bpf_program__unload
- [ ] bpf_program__attach
- [ ] bpf_program__attach_perf_event
- [ ] bpf_program__attach_perf_event_opts
- [ ] bpf_program__attach_kprobe
- [ ] bpf_program__attach_kprobe_opts
- [ ] bpf_program__attach_uprobe
- [ ] bpf_program__attach_uprobe_opts
- [ ] bpf_program__attach_tracepoint
- [ ] bpf_program__attach_tracepoint_opts
- [ ] bpf_program__attach_raw_tracepoint
- [ ] bpf_program__attach_trace
- [ ] bpf_program__attach_lsm
- [ ] bpf_program__attach_cgroup
- [ ] bpf_program__attach_netns
- [ ] bpf_program__attach_xdp
- [ ] bpf_program__attach_freplace
- [ ] bpf_program__attach_iter
- [ ] bpf_program__set_socket_filter
- [ ] bpf_program__set_tracepoint
- [ ] bpf_program__set_raw_tracepoint
- [ ] bpf_program__set_kprobe
- [ ] bpf_program__set_lsm
- [ ] bpf_program__set_sched_cls
- [ ] bpf_program__set_sched_act
- [ ] bpf_program__set_xdp
- [ ] bpf_program__set_perf_event
- [ ] bpf_program__set_tracing
- [ ] bpf_program__set_struct_ops
- [ ] bpf_program__set_extension
- [ ] bpf_program__set_sk_lookup
- [ ] bpf_program__get_type
- [ ] bpf_program__set_type
- [ ] bpf_program__get_expected_attach_type
- [ ] bpf_program__set_expected_attach_type
- [ ] bpf_program__set_attach_target
- [ ] bpf_program__is_socket_filter
- [ ] bpf_program__is_tracepoint
- [ ] bpf_program__is_raw_tracepoint
- [ ] bpf_program__is_kprobe
- [ ] bpf_program__is_lsm
[...]

PR #335, opened 2023-06-16
Wrap `bpf_prog_attach` and `bpf_prog_detach2`

mozillazg:
https://github.com/aquasecurity/libbpfgo/issues/91

### DIFF
--- a/libbpfgo.go
+++ b/libbpfgo.go
@@ -1711,6 +1711,35 @@ func doAttachUprobe(prog *BPFProg, isUretprobe bool, pid int, path string, offse
 	return bpfLink, nil
 }
 
+type AttachFlag uint32
+
+const (
+	BPFFNone          AttachFlag = 0
+	BPFFAllowOverride AttachFlag = C.BPF_F_ALLOW_OVERRIDE
+	BPFFAllowMulti    AttachFlag = C.BPF_F_ALLOW_MULTI
+	BPFFReplace       AttachFlag = C.BPF_F_REPLACE
+)
+
+// AttachGenericFD attaches the BPFProgram to a targetFd at the specified attachType hook.
+func (p *BPFProg) AttachGenericFD(targetFd int, attachType BPFAttachType, flags AttachFlag) error {
+	progFd := C.bpf_program__fd(p.prog)
+	errC := C.bpf_prog_attach(progFd, C.int(targetFd), C.enum_bpf_attach_type(int(attachType)), C.uint(uint(flags)))
+	if errC < 0 {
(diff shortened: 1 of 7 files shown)


libbpfgo.go
```diff
@@ -1711,6 +1711,35 @@ func doAttachUprobe(prog *BPFProg, isUretprobe bool, pid int, path string, offse
...
+
+const (
+	BPFFNone          AttachFlag = 0
+	BPFFAllowOverride AttachFlag = C.BPF_F_ALLOW_OVERRIDE
+	BPFFAllowMulti    AttachFlag = C.BPF_F_ALLOW_MULTI
+	BPFFReplace       AttachFlag = C.BPF_F_REPLACE
+)
+
+// Attach attaches the BPFProgram to a targetFd at the specified attachType hook.
+func (p *BPFProg) Attach(targetFd int, attachType BPFAttachType, flags AttachFlag) error {
```
rafaeldtinoco, line 1724: We have `AttachGeneric` which is similar, but does not allow to specify a fd to attach to. Should we name this `AttachGenericFD`?

rafaeldtinoco: I also see that you're using it for attaching a program to a cgroupFD (Which could be done by the attach_cgroup method, no ?). I want to understand if:

- should we make attachgeneric generic enough to allow a fd to attach to ? (or is it just the cgroup case and then we already have an attachment function for it that could be used and/or changed)

- this is a specific need for something else (then we should either change AttachGeneric OR the selftest to differentiate between the cgroup attachment).


WDYT ?

mozillazg: > We have `AttachGeneric` which is similar, but does not allow to specify a fd to attach to. Should we name this `AttachGenericFD`?

SGTM. I'll rename it to `AttachGenericFD`.

> I also see that you're using it for attaching a program to a cgroupFD (Which could be done by the attach_cgroup method, no ?)

That program can't be attached by the attach_cgroup method.

> should we make attachgeneric generic enough to allow a fd to attach to ? (or is it just the cgroup case and then we already have an attachment function for it that could be used and/or changed)

That would be a breaking change and would make the attachgeneric method very complex. we should avoid it.

> this is a specific need for something else (then we should either change AttachGeneric OR the selftest to differentiate between the cgroup attachment).

Yes, This is a specific method for attaching cgroup, sockops, flow_dissector, lirc_mode2, and sockmap programs. 
I'll change the selftest to attach a sockmap program.

mozillazg:
@rafaeldtinoco ping~

rafaeldtinoco:
Thanks for your contribution, as always.

PR #338, opened 2023-07-10
feat: add GetMapByName

aymericDD:
Currently, the library allows you to get a map from an **eBPF** program loaded by themselves. If a map is already created, you cannot get it. This function aims to get a map already created.


.gitignore
```diff
@@ -7,3 +7,122 @@ selftest/uprobe/ctest
...
+.idea/**/tasks.xml
+.idea/**/usage.statistics.xml
+.idea/**/dictionaries
+.idea/**/shelf
+
+# AWS User-specific
+.idea/**/aws.xml
+
+# Generated files
+.idea/**/contentModel.xml
```
rafaeldtinoco, line 29: Could you please remove all changes to .gitignore ?

selftest/get-map-by-name/run.sh
rafaeldtinoco, line 1: This file is just a symlink in the other examples, I believe that can be the same here, right ?

selftest/get-map-by-name/Makefile
rafaeldtinoco, line 1: This file is just a symlink in the other selftests, can be the same here, right ?

libbpfgo.go
```diff
@@ -531,6 +531,47 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
 	}, nil
 }
 
+/*
+GetMapByName get an existing map from userspace.
```
rafaeldtinoco, line 535: Could you use // in the function godoc like the other ones ?

nit: "GetMapByName gets an existing map..." (verb)
nit: "can be used for populating BPF..." (verb)
nit: "... or hasH of maps" (typo)

libbpfgo.go
```diff
@@ -531,6 +531,47 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
...
+		if mapName != name {
+			continue
+		}
+
+		return &BPFMap{
+			name: name,
+			fd:   fd,
+		}, nil
+	}
+	return nil, fmt.Errorf("the %s map does not exists", name)
```
rafaeldtinoco, line 572: nit: "does not exist" (typo on the verb)

rafaeldtinoco:
image

Selftest is also failing, could you please address those issues ? Thanks for your contribution, I'll take a better look on the next iteration. Cheers!

rafaeldtinoco (changes requested):
image

Selftest is also failing, could you please address those issues ? Thanks for your contribution, I'll take a better look on the next iteration. Cheers!

aymericDD:
Sorry for the convenience... I hope the test will works 🙏

aymericDD:
Any updates? 🔍

rafaeldtinoco:
> Any updates? 🔍

Sorry for the delay, been out sick, slowly catching up on things. Will check this soon.

aymericDD:
Thanks! 👍  No problem! I hope you feel better 🙏 Cheers!

libbpfgo.go
```diff
@@ -531,6 +531,45 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
 	}, nil
 }
 
+// GetMapByName gets an existing map
+// Can be used for populating BPF array of maps or hash of maps. Currently, the returned map can only be updated.
+func GetMapByName(name string) (*BPFMap, error) {
```
geyslan, line 536: The logic used in GetMapByName is also being worked on in #343. @aymericDD could you please hold this effort for a while? So you can rebase it on top of upcoming changes. Thank you.

aymericDD: Yes, no problem 👍 We manage to not be dependent of the libbpfgo for now, but it will more clean to be able to use it for the future.

geyslan: @aymericDD we are going in a different direction when using map of maps. You can see more in https://github.com/aquasecurity/libbpfgo/pull/343#issuecomment-1660998211.

That said, for your GetMapByName() to be in the managed part of the lib, it must have the BPFModule as an argument and search the map in it.

Would that solve your use case?

aymericDD: Thanks to let me know 👍 
FWIK: it does not solve my use case, because the eBPF map is loaded from an [eBPF tc filter](https://man7.org/linux/man-pages/man8/tc-bpf.8.html). If I try to load a `BPFModule` `bpfModule, err := bpf.NewModuleFromFile("myprgm.bpf.o")`, it will create a new map not related to the map of the eBPF tc filter. The idea of this function is really to get a map already instantiated from another program. I hope it makes sense 👍  

The correct behavior:
image

If I use the BPF module:
image

geyslan: I see, you need access to maps instantiated by other origins (not managed by any modules loaded through conventional means).

This falls in the case we cannot return a `BPFMap` type that will fail at some point of execution (for example, through methods that need a reference to the libbpf module and/or map objects - `b.bpfMap`).

In my previous PR https://github.com/aquasecurity/libbpfgo/pull/343, which was a continuation of @kakkoyun's, I tried to work around this issue by putting checks on all BPFMap methods, but in hindsight we assumed that it's not the right approach.

We need to make this obvious to the libbpfgo consumer, so BFPMapOfMaps is a soon-to-be-coming type with its own possible (and necessary) methods.

Regarding your use case, I assume that we need to return another type of BPFMap than the managed one and handle all its methods without confusing the latter. `BPFMapUnmanaged`? I'm still terrible at naming things.

libbpfgo.go
```diff
@@ -531,6 +531,45 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
...
+		infolen := C.uint(unsafe.Sizeof(info))
+		err = C.bpf_obj_get_info_by_fd(fd, unsafe.Pointer(&info), &infolen)
+		if err != 0 {
+			return nil, fmt.Errorf("could not get the map info: %w", syscall.Errno(-err))
+		}
+
+		mapName := C.GoString((*C.char)(unsafe.Pointer(&info.name[0])))
+		if mapName != name {
+			continue
+		}
```
geyslan, line 563: And to make sure we're dealing with the right BPF map, I assume we must check the pin path, instead of its name. There might be different maps with the same name (from different BPF objects/origins).

image

libbpfgo.go
aymericDD: Exactly 👌  My problem is the map is instantiated by another origin. The idea of having a new kind a map looks good to me 👍 For the name of the map it's up to you 😉

libbpfgo.go
aymericDD: Yes, it is a problem 🤔 How can I get the bin path of the map? We can also return an error like `MultipleMapFoundsError` or a list of maps.

geyslan: > How can I get the bin path of the map?

Libbpf holds that information, but we're not treating with libbpf generated maps.

The pinning sets `attr.pathname`. But further, retrieving map information, unfortunately bpf_map_info has no `filepath` field.

image

There's another issue with the pin path approach: maps from other origin may not have been pinned.

> We can also return an error like `MultipleMapFoundsError` or a list of maps.

Maybe return a slice of map IDs, so the caller would decide which to choose based on later access to the respective information (type, key/value size, flags, etc.).

geyslan: > It looks good 👍 I can update the PR to return a list of Maps rather the first map matched the name criteria.

Excellent. I would just ask you to wait a bit. I'm doing local testing and looking at splitting maps managed by libbpf and user-created ones (or from another origin) to align in the API.

geyslan: @aymericDD sorry about the delay.

We have merged major map changes, which provide helpers for your attempt. The logic should be in `map-low.go` as that is where the low-level API for maps resides. As agreed, `GetMapsByName()` should return a slice of map IDs.

libbpfgo.go
```diff
@@ -531,6 +531,45 @@ func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries i
...
+
+	for {
+		err := C.bpf_map_get_next_id(startId, &nextId)
+		if err != 0 {
+			return nil, fmt.Errorf("could not get the map: %w", syscall.Errno(-err))
+		}
+
+		startId = nextId + 1
+
+		fd := C.bpf_map_get_fd_by_id(nextId)
```
geyslan, line 548: Use `GetMapFDByID`

libbpfgo.go
aymericDD: Nice! I will take a look 👍

aymericDD:
Is it possible to release a new version of this library? It will help me 🙏

geyslan:
> Is it possible to release a new version of this library? It will help me 🙏

It's being baked. It'll be available asap.

geyslan:
@aymericDD https://github.com/aquasecurity/libbpfgo/releases/tag/v0.5.0-libbpf-1.2

aymericDD:
Thanks @geyslan
After updating my vendor I have the following error when I try to compile the program:
`cgo: ../vendor/github.com/aquasecurity/libbpfgo/libbpfgo.go:1077:18: unexpected: -1-byte enum type - enum bpf_cgroup_iter_order {}`

geyslan:
@aymericDD I haven't detected that error (yet).

image

image

Could you try to change that assignment to `cgroupIterOrderC := C.enum_bpf_cgroup_iter_order(opts.CgroupIterOrder)` and check it again.

aymericDD:
Thanks for your help. I am still debugging, but I have a question. Which version of the kernel are you using? I am using the `v5.15.x`, but I supposed you use `v6.x`.

geyslan:
> Which version of the kernel are you using? I am using the `v5.15.x`, but I supposed you use `v6.x`.

I'm testing it on the amd64-vm (Vagrantfile) which uses `5.15.0-78-generic` - without errors.

Note that the only use of `AttachIter` in selftests is passing a zeroed value `IterOpts`: https://github.com/aquasecurity/libbpfgo/blob/c11365cbf2c67619350e72e3f3c67e01082780fb/selftest/iter/main.go#L39

PR #341, opened 2023-07-20
fix: vagrant image install

geyslan:
linux-tools-generic install was failing on generic/ubuntu2204 image.

- Change amd64 to ubuntu/jammy64.
- Change arm64 to bento/ubuntu-22.04-arm64
- Remove ssh workaround setting (not needed for jammy64).


rafaeldtinoco (approved):
LGTM

PR #345, opened 2023-08-01
helpers: handle non numeric kernel versions

NDStrahilevitz:
Some kernels report their version with non numeric characters such as +. Handle these cases in kernel version comparisons.


geyslan (approved):
LGTM

PR #348, opened 2023-08-01
osinfo: support additional alma linux id

NDStrahilevitz:
Some alma distributions use almalinux instead of alma as the ID field in os-release.


geyslan (approved):
LGTM

PR #350, opened 2023-08-02
fix(selftest): handle arm64 function names

geyslan:
commit e7932792aaff5ada7eb29232a03563660a54cc1b

    chore(selftest): lower ping timeout to 10 secs
    
    ping command timeout was higher than the iter selftest timeout.

commit 46d619adc84c1ffd0a88593437e11ff3cbe15666

    fix(selftest): handle arm64 function names
    
    Check current architecture and use the correct function names.
    
    Despite the fact that selftests are now handling the right function
    names for arm64, some are still not working due to the fact that arm64
    still does not support generic attaching `bpf_program__attach()`.
    
    Error:
    
    libbpf: prog 'mmap_fentry': failed to attach: ERROR: strerror_r(-524)=22
    failed to attach program: errno 524
    [!] ERROR: test error
    make[1]: *** [Makefile:81: run-static] Error 4
    make[1]: Leaving directory '/vagrant/selftest/percpu'
    
    '#define ENOTSUPP       524     /* Operation is not supported */'
    https://elixir.bootlin.com/linux/latest/source/include/linux/errno.h#L27

PR #349, opened 2023-08-02
chore: increase timeout for selftests

geyslan:
commit e93ed57db2c046d529c219004dec228f7e21473d

    chore: increase timeout for selftests

PR #364, opened 2023-08-08
fix(selftest): use channel for sync

geyslan:
Context:
https://github.com/aquasecurity/libbpfgo/pull/349#issuecomment-1662157878
https://github.com/aquasecurity/libbpfgo/pull/350#issuecomment-1662991998

fix(selftest): use channel for sync
    
    attachgenericfd and perfbuffers selftests were using a goroutine without
    any syncronization mechanism.

### DIFF
--- a/selftest/attachgenericfd/main.go
+++ b/selftest/attachgenericfd/main.go
@@ -3,9 +3,11 @@ package main
 import "C"
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


geyslan:
> This is being hard to catch since I can't reproduce it locally.

It seems solved.

image

PR #368, opened 2023-08-24
fix(map): fix error wrap

geyslan:
Error wrapping verb can be used only one time per error.


geyslan:
Merging ignoring the suggestion. It'll be tackled in a further PR.

PR #374, opened 2023-08-31
Map batch API

geyslan:
Close: https://github.com/aquasecurity/libbpfgo/issues/162

commit 9382e51652377376ef9459901c49abbcf39758b3 (HEAD -> map-batch, origin/map-batch)

    fix(map)!: fix map batch methods
    
    This breaks the API, returning a count of the number of elements
    processed for each batch method.
    
    It also fixes the check against ENOENT and E2BIG.
    
    The selftest has been updated to reflect the new API.

### DIFF
--- a/map-low.go
+++ b/map-low.go
@@ -318,10 +318,12 @@ func (m *BPFMapLow) DeleteKey(key unsafe.Pointer) error {
 // BPFMapLow Batch Operations
 //
 
-func (m *BPFMapLow) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, error) {
+// GetValueBatch gets the values with the given keys from the map.
+// It returns the values and the number of read elements.
+func (m *BPFMapLow) GetValueBatch(keys, startKey, nextKey unsafe.Pointer, count uint32) ([][]byte, uint32, error) {
 	valueSize, err := calcMapValueSize(m.ValueSize(), m.Type())
 	if err != nil {
-		return nil, fmt.Errorf("map %s %w", m.Name(), err)
+		return nil, 0, fmt.Errorf("map %s %w", m.Name(), err)
 	}
 
 	var (
@@ -332,7 +334,7 @@ func (m *BPFMapLow) GetValueBatch(keys unsafe.Pointer, startKey, nextKey unsafe.
 
 	optsC, errno := C.cgo_bpf_map_batch_opts_new(C.BPF_ANY, C.BPF_ANY)
 	if optsC == nil {
-		return nil, fmt.Errorf("failed to create bpf_map_batch_opts: %w", errno)
+		return nil, 0, fmt.Errorf("failed to create bpf_map_batch_opts: %w", errno)
 	}
 	defer C.cgo_bpf_map_batch_opts_free(optsC)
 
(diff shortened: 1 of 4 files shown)


geyslan:
@rafaeldtinoco I got the other refactors in (were getting dust), but this breaks API.

kakkoyun:
@rafaeldtinoco @geyslan We would love to be able to use batch APIs downstream @parca-dev.
How could we be helpful to push this thorough? 

cc @javierhonduco

kakkoyun:
> Hey @kakkoyun and @javierhonduco. We would love to get your review on this PR.

Sure thing! 👍

map-low.go
```diff
@@ -416,22 +423,23 @@ func (m *BPFMapLow) UpdateBatch(keys, values unsafe.Pointer, count uint32) error
 		optsC,
 	)
 	errno = syscall.Errno(-retC)
-	if retC < 0 {
-		if errno != syscall.EFAULT && uint32(countC) != count {
-			return fmt.Errorf("failed to update ALL elements in map %s, updated (%d/%d): %w", m.Name(), uint32(countC), count, errno)
-		}
-		return fmt.Errorf("failed to batch update elements in map %s: %w", m.Name(), errno)
+	// retC < 0 && errno == syscall.E2BIG indicates a partial update.
+	if retC < 0 && (errno != syscall.E2BIG || countC == 0) {
```
javierhonduco, line 427: Ohh good catch, I think I introduced this bug. I was trying to find where in the kernel E2BIG is returned if there's a partial update and couldn't find that code path. Would you mind sharing it? Thanks!!

geyslan: I've followed this path:

https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L5401
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L4883
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L4845
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/arraymap.c#L764
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L1750
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/syscall.c#L183
https://elixir.bootlin.com/linux/latest/source/kernel/bpf/arraymap.c#L853

But the initial hint was the return types from sys_bpf: https://man7.org/linux/man-pages/man2/bpf.2.html#ERRORS

```
ERRORS
       E2BIG  The eBPF program is too large or a map reached the max_entries limit (maximum
              number of elements).
```

javierhonduco: Thanks for the context, I hadn't checked array maps where perhaps this is more obvious. Should we then augment the tests to check that the partial / total batch operations work well with other map types such as arrays?

map-low.go
```diff
@@ -416,22 +423,23 @@ func (m *BPFMapLow) UpdateBatch(keys, values unsafe.Pointer, count uint32) error
 		optsC,
 	)
 	errno = syscall.Errno(-retC)
-	if retC < 0 {
-		if errno != syscall.EFAULT && uint32(countC) != count {
-			return fmt.Errorf("failed to update ALL elements in map %s, updated (%d/%d): %w", m.Name(), uint32(countC), count, errno)
-		}
-		return fmt.Errorf("failed to batch update elements in map %s: %w", m.Name(), errno)
+	// retC < 0 && errno == syscall.E2BIG indicates a partial update.
```
javierhonduco, line 426: Is the comment out of date? (`!=` rather than `==`)?

geyslan: I don't think so, it states that it can be updated partially until getting E2BIG. Perhaps it can be better explained. WDYT?

javierhonduco: It's mostly a nit, but I find it confusing that this comment is here when it's dealing with another case. But again, it's a personal style preference, I am not super strong about this one!

geyslan: Possibly it would be better as:

```c
// errno == syscall.E2BIG may indicate a partial update, so only consider it as an error when (errno != syscall.E2BIG || countC == 0).
```

The same pattern is valid to the other function comments. WDYT?

geyslan:
Sorry for the delay on this. Going to merge as it is, any required changes are welcome.

PR #377, opened 2023-09-12
feat: GetMapsIdsByName

aymericDD:
Currently, the library allows you to get a map from an eBPF program loaded by themselves. If a map is already created, you cannot get it. This function aims to get a list of maps Ids by name already created.


libbpfgo.go
```diff
@@ -393,6 +393,34 @@ func (m *Module) InitGlobalVariable(name string, value interface{}) error {
 	return err
 }
 
+// GetMapsByName retrieves BPF maps with the specified name.
+func GetMapsByName(name string) []*BPFMapLow {
```
geyslan, line 397: I see a problem with returning ready-made BPFMapLow instead of map IDs: the caller might simply forget to close the maps' FDs and cause them to leak. I think forcing the caller to call `GetMapByID` for each map ID makes the caller's responsibility for management crystal.

`func GetMapsByName(name string) []int {`

WDYT?

libbpfgo.go
```diff
@@ -393,6 +393,34 @@ func (m *Module) InitGlobalVariable(name string, value interface{}) error {
...
+
+		startId = nextId + 1
+
+		bpfMapLow, errMap := GetMapByID(uint32(nextId))
+		if errMap != nil {
+			return bpfMaps
+		}
+
+		if bpfMapLow.Name() != name {
+			continue
```
geyslan, line 417: Here we need to close the map FD of the unmatch.

libbpfgo.go
```diff
@@ -393,6 +393,34 @@ func (m *Module) InitGlobalVariable(name string, value interface{}) error {
 	return err
 }
 
+// GetMapsByName retrieves BPF maps with the specified name.
+func GetMapsByName(name string) []*BPFMapLow {
+	bpfMaps := []*BPFMapLow{}
```
geyslan, line 398:
```suggestion
	bpfMapsIDs := []int{}
```

selftest/map-get-maps-by-name/main.bpf.c
```diff
@@ -0,0 +1,20 @@
...
+    __uint(value_size, sizeof(u32));
+} test SEC(".maps");
+
+SEC("kprobe/sys_execve")
+int kprobe__sys_execve(struct pt_regs *ctx)
+{
+    return 0;
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
```
geyslan, line 20: nit: add an end new line.

image

geyslan (commented):
@aymericDD thank you again for your grand contribution!

libbpfgo.go
aymericDD: Sorry, I was not aware of the closing of maps' FDs. In this case, yes, it is a good idea to just return Ids and let the user handle the rest.

geyslan:
Closing this in favour of #378. @aymericDD I would appreciate your review. Thanks.

PR #384, opened 2023-10-10
Simplify GetMapsIDsByName

aymericDD:
The `id` is incremented by `C.bpf_map_get_next_id` so no need to define the `startId` and `nextId`. I encounter some weird behavior in my project when I was using map with a BPF_MAP_TYPE_HASH type.

### DIFF
--- a/map-low.go
+++ b/map-low.go
@@ -37,29 +37,6 @@ type BPFMapCreateOpts struct {
 	MapIfIndex            uint32
 }
 
-// GetMapNextIDENOENTError is a custom error type that represents the "no entry" error
-// when attempting to get the next map ID.
-type GetMapNextIDENOENTError struct{}
-
-// GetMapNextIDEPERMError is a custom error type that represents the "permission denied" error
-// when attempting to get the next map ID.
-type GetMapNextIDEPERMError struct{}
-
-// GetMapNextIDEINVALError is a custom error type that represents the "invalid value" error
-// when attempting to get the next map ID.
-type GetMapNextIDEINVALError struct{}
-
-func (g GetMapNextIDENOENTError) Error() string {
-	return fmt.Sprintf("Error no entry")
-}
-
-func (g GetMapNextIDEPERMError) Error() string {
-	return fmt.Sprintf("Error permission denied")
-}
-func (g GetMapNextIDEINVALError) Error() string {
-	return fmt.Sprintf("Error invalid value")
-}
-
 // CreateMap creates a new BPF map with the given parameters.
 func CreateMap(mapType MapType, mapName string, keySize, valueSize, maxEntries int, opts *BPFMapCreateOpts) (*BPFMapLow, error) {
 	mapNameC := C.CString(mapName)
@@ -125,24 +102,14 @@ func GetMapByID(id uint32) (*BPFMapLow, error) {
 
 // GetMapNextID retrieves the next available map ID after the given startID.
 // It returns the next map ID and an error if one occurs during the operation.
-func GetMapNextID(id uint32) (uint32, error) {
-	startID := C.uint(id)
-	retC := C.bpf_map_get_next_id(startID, &startID)
+func GetMapNextID(startId uint32) (uint32, error) {
+	startIDC := C.uint(startId)
+	retC := C.bpf_map_get_next_id(startIDC, &startIDC)
 	if retC == 0 {
-		return uint32(startID), nil
+		return uint32(startIDC), nil
 	}
 
-	errno := syscall.Errno(-retC)
-	switch errno {
-	case syscall.ENOENT:
-		return uint32(startID), GetMapNextIDENOENTError{}
-	case syscall.EINVAL:
-		return uint32(startID), GetMapNextIDEINVALError{}
-	case syscall.EPERM:
-		return uint32(startID), GetMapNextIDEPERMError{}
-	default:
-		return uint32(startID), fmt.Errorf("failed to get next map id: %w", errno)
-	}
+	return uint32(startIDC), fmt.Errorf("failed to get next map id: %w", syscall.Errno(-retC))
 }
 
 // GetMapsIDsByName searches for maps with a specified name and collects their IDs.
@@ -179,7 +146,7 @@ func GetMapsIDsByName(name string, startId uint32) ([]uint32, error) {
 	for {
 		startId, err = GetMapNextID(startId)
 		if err != nil {
-			if errors.Is(err, GetMapNextIDENOENTError{}) {
+			if errors.Is(err, syscall.ENOENT) {
 				return bpfMapsIds, nil
 			}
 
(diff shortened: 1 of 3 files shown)


geyslan:
Hello @aymericDD. Could you please explain this weird behaviour so we can provide more context for this change.

aymericDD:
When I was using a **BPF_MAP_TYPE_HASH** with a struct key and struct value, the map was not returned in the list of IDs. I take a look at the [bpftool](https://github.com/libbpf/bpftool/blob/master/src/map.c#L673) and I notice the ID was not re-incremented manually.

map-low.go
```diff
@@ -105,11 +105,10 @@ func GetMapByID(id uint32) (*BPFMapLow, error) {
 func GetMapsIDsByName(name string) ([]uint32, error) {
 	bpfMapsIds := []uint32{}
 
-	startId := C.uint(0)
-	nextId := C.uint(0)
+	id := C.uint(0)
 
 	for {
-		retC := C.bpf_map_get_next_id(startId, &nextId)
+		retC := C.bpf_map_get_next_id(id, &id)
```
geyslan, line 111: I suppose passing the same object isn't a problem, since the libbpf helper only writes to it after reading it and when no error occurred. 👍🏼

But the problem you faced must be related to an error other than ENOENT. WDYT?

aymericDD: **FWIK** the `GetMapsIDsByName` does not return an `error`. You can try locally by rollbacking the `GetMapsIDsByName` _(with StartId end EndId)_, and normally the test should fail because no IDs are returned. The error should be: `the test_hash_name map should be found`

geyslan: Sorry for the delay.

https://elixir.bootlin.com/linux/v6.5.7/source/kernel/bpf/syscall.c#L3778

`bpf_obj_get_next_id()` has four points of error return. We're considering only `ENOENT` in `GetMapsIDsByName()` what can cause a infinite loop when the error is one other.

In hindsight I assume that this wrapping is not one of the best choices, since we should abort it at any error returned but we don't provide a way to restart the fetching based on the last id successfully got.

My proposal is to create a proper wrapper getting only the next map id, using the same signature or similar like:

`GetMapNextID(startID uint32, nextID *uint32) error`

or

`GetMapNextID(startID uint32) (uint32, error)`

And convert `GetMapsIDsByName` to make use of GetMapNextID. In it consider all the types of errors, returning the filled slice of IDs (partial or not) and the error (whenever it occurs).

`GetMapsIDsByName(name string, startID uint32) ([]uint32, error)`

The GetMapsIDsByName caller would check for ENOENT, if it's, all maps IDs with that name would be returned. If it's a different kind, it's up to the user to try it again using the last ID from the []uint32. WDYT? Would you like to continue with these changes?

aymericDD: I agree with you 👍 Handling all kinds of error could be nice and having a wrapper for `GetMapNexId` is a nice idea. But in the scenario I don't notice any error returned. Just the map is empty.

aymericDD: > bpf_obj_get_next_id() has four points of error return. We're considering only ENOENT in GetMapsIDsByName() what can cause a infinite loop when the error is one other.

I don't think we are in an infinity loop if an error is different than ENOENT, because we wrap all other errors from `bpf_map_get_next_id`:

``` go
		retC := C.bpf_map_get_next_id(id, &id)
		errno := syscall.Errno(-retC)
		if retC < 0 {
			if errno == syscall.ENOENT {
				return bpfMapsIds, nil
			}

			return bpfMapsIds, fmt.Errorf("failed to get next map id: %w", errno)
		}
```

aymericDD: For the GetMapNextId something like that could do the job? :
``` go

func GetMapNextID(startID uint32) (uint32, error) {
	retC := C.bpf_map_get_next_id(startID, &startID)
	errno := syscall.Errno(-retC)
	if retC < 0 {
		switch errno {
		case syscall.ENOENT:
			return startID, GetMapNextIDENOENTError{}
		case syscall.EINVAL:
			return startID, GetMapNextIDEINVALError{}
		case syscall.EPERM:
			return startID, GetMapNextIDEPERMError{}
		default:
			return startID, fmt.Errorf("failed to get next map id: %w", errno)
		}
	}

	return startID, nil
}
```

geyslan: > I don't think we are in an infinity loop if an error is different than ENOENT, because we wrap all other errors from bpf_map_get_next_id:

You're right. 👍🏼

map-low.go
```diff
@@ -99,29 +123,70 @@ func GetMapByID(id uint32) (*BPFMapLow, error) {
...
+	switch errno {
+	case syscall.ENOENT:
+		return uint32(startID), GetMapNextIDENOENTError{}
+	case syscall.EINVAL:
+		return uint32(startID), GetMapNextIDEINVALError{}
+	case syscall.EPERM:
+		return uint32(startID), GetMapNextIDEPERMError{}
+	default:
+		return uint32(startID), fmt.Errorf("failed to get next map id: %w", errno)
+	}
```
geyslan, line 145: I realized that creating our own error types would increase maintenance efforts and ignore the original error message. So how about just returning the original encapsulated error? We know they are syscall.EXX, so the caller can compare them directly. WDYT?

```suggestion
	return uint32(startID), fmt.Errorf("failed to get next map id: %w", errno)
```

map-low.go
```diff
@@ -99,29 +123,70 @@ func GetMapByID(id uint32) (*BPFMapLow, error) {
...
+	)
 
 	for {
-		retC := C.bpf_map_get_next_id(startId, &nextId)
-		errno := syscall.Errno(-retC)
-		if retC < 0 {
-			if errno == syscall.ENOENT {
+		startId, err = GetMapNextID(startId)
+		if err != nil {
+			if errors.Is(err, GetMapNextIDENOENTError{}) {
```
geyslan, line 182: As commented above.

```suggestion
			if errors.Is(err, syscall.ENOENT) {
```

map-low.go
```diff
@@ -99,29 +100,60 @@ func GetMapByID(id uint32) (*BPFMapLow, error) {
...
+//
+//	var mapIDs []uint32      // Initialize an empty slice to collect map IDs.
+//	var err error            // Initialize an error variable.
+//
+//	// Retry mechanism in case of errors using the last processed 'startId'.
+//	for {
+//	    mapIDs, err = GetMapsIDsByName(name, startId)
+//	    if err != nil {
+//	        // Handle other errors, possibly with a retry mechanism.
+//	        // You can use the 'startId' who contains the last processed map ID to continue the search.
```
geyslan, line 134: I didn't see this. `startId` is not a pointer to be changed. The user can only rely on the last element of the returned bpfMapsIds.

aymericDD: 🙈 Indeed! Well spotted!

geyslan:
@aymericDD is it waiting review?

aymericDD:
Yes 👍

geyslan (approved):
LGTM

geyslan:
@aymericDD thanks again for your contribution!

PR #385, opened 2023-10-16
Feature/parse gup and vm flags

AlonZivony:
Add flags to parse GUP and VM flags arguments.

### DIFF
--- a/helpers/argumentParsers.go
+++ b/helpers/argumentParsers.go
@@ -3062,3 +3062,98 @@ func ParseGUPFlags(rawValue uint64) GUPFlag {
 
 	return GUPFlag{stringValue: strings.Join(f, "|"), rawValue: uint32(rawValue)}
 }
+
+// =====================================================
+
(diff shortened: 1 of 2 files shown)


helpers/argumentParsers.go
```diff
@@ -2969,3 +2969,317 @@ func ParseIoUringRequestFlags(rawValue uint64) IoUringRequestFlag {
 
 	return IoUringRequestFlag{stringValue: strings.Join(f, "|"), rawValue: uint32(rawValue)}
 }
+
+// =====================================================
+
+// GupFlag represents GUP (Get User Pages) flags since version 6.3
+type GupFlag struct {
```
geyslan, line 2976: golang nit: `GUPFlag` as the letters are initials of different words. P.S.: Hopefully it's not the case of the ID != Id conundrum 😄.

geyslan (approved):
LGTM, just a nit, WDYT?

PR #408, opened 2024-03-13
Added parsing for fsnotify mask and object type

oshaked1:
The fsnotifty mask consists of flags that specify which filesystem events should be monitored. They are used internally by the fsnotify system and correspond to similarly named flags used by the individual APIs that use fsnotify (dnotify, inotify and fanotify).


geyslan (approved):
LGTM

PR #409, opened 2024-03-18
feature: Add AttachProgs/DetachProgs api for bpf object

chentao-kernel:
Add AttachProgs/DetachProgs api for bpf object like api in libbpf: bpf_object__attach_skeleton/bpf_object__detach_skeleton. So we don't need to specify the program name to attach one by one.

### DIFF
--- a/module.go
+++ b/module.go
@@ -391,3 +391,65 @@ func (m *Module) Iterator() *BPFObjectIterator {
 		prevMap:  nil,
 	}
 }
+
+func (m *Module) linkExist(prog *BPFProg) bool {
+	for _, link := range m.links {
+		if link.prog.Name() == prog.Name() {
+			return true
+		}
+	}
+
+	return false
+}
+
+// AttachPrograms attach all loaded and no attached progs once like bpf_object__attach_skeleton
+func (m *Module) AttachPrograms() error {
+	iters := m.Iterator()
+	for {
+		prog := iters.NextProgram()
+		if prog == nil {
+			break
+		}
+
+		if !prog.Autoload() || !prog.Autoattach() {
+			continue
+		}
+		// if link already exist (is attached), skip it
+		if m.linkExist(prog) {
+			continue
+		}
+
+		link, err := prog.AttachGeneric()
+		if err != nil {
+			return err
+		}
+
+		m.links = append(m.links, link)
+	}
+
+	return nil
+}
+
+// DetachPrograms detach all attached progs once like bpf_object__detach_skeleton
+func (m *Module) DetachPrograms() error {
+	errInfo := make(map[string]error)
+
+	for _, link := range m.links {
+		err := link.Destroy()
+		if err != nil {
+			errInfo[link.prog.Name()] = err
+		}
+	}
+	m.links = nil
+
+	if len(errInfo) > 0 {
+		var str string
+		for name, err := range errInfo {
+			str += fmt.Sprintf(" [prog:%s,err:%s]", name, err)
+		}
+
+		return fmt.Errorf("link destroy failed:%s", str)
+	}
+
+	return nil
+}
--- a/selftest/module-attach-detach/main.bpf.c
+++ b/selftest/module-attach-detach/main.bpf.c
@@ -1,0 +1,24 @@
+//+build ignore
+
+#include <vmlinux.h>
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#ifdef __TARGET_ARCH_amd64
+SEC("fentry/__x64_sys_mmap")
+#elif defined(__TARGET_ARCH_arm64)
+SEC("fentry/__arm64_sys_mmap")
+#endif
+int sys_mmap(struct pt_regs *ctx)
+{
+    return 0;
+}
+
+SEC("kprobe/do_sys_open")
+int kprobe__sys_open(struct pt_regs *ctx)
+{
+    return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/prog.go
+++ b/prog.go
@@ -111,6 +111,10 @@ func (p *BPFProg) SetAutoload(autoload bool) error {
 	return nil
 }
 
+func (p *BPFProg) Autoload() bool {
+	return bool(C.bpf_program__autoload(p.prog))
+}
+
 func (p *BPFProg) SetAutoattach(autoload bool) {
 	C.bpf_program__set_autoattach(p.prog, C.bool(autoload))
 }
--- a/selftest/module-attach-detach/Makefile
+++ b/selftest/module-attach-detach/Makefile
@@ -1,0 +1,1 @@
+../common/Makefile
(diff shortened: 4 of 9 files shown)


module.go
```diff
@@ -391,3 +391,33 @@ func (m *Module) Iterator() *BPFObjectIterator {
...
 }
+
+func (m *Module) AttachProgs() error {
+	iters := m.Iterator()
+	for {
+		prog := iters.NextProgram()
+		if prog == nil {
+			break
+		}
+		link, err := prog.AttachGeneric()
```
geyslan, line 402: Before attempting to attach, shouldn't we check for autoload and autoattach as bpf_object__attach_skeleton does:

```c
		if (!prog->autoload || !prog->autoattach)
			continue;
```

And if the link is already set?

```
		/* if user already set the link manually, don't attempt auto-attach */
		if (*link)
			continue;
```

module.go
```diff
@@ -391,3 +391,33 @@ func (m *Module) Iterator() *BPFObjectIterator {
...
+		if err != nil {
+			return err
+		}
+		m.links = append(m.links, link)
+	}
+
+	return nil
+}
+
+func (m *Module) DetachProgs() error {
```
geyslan, line 412: Shouldn't we remove the link from `m.links` when successfully detroyed?

module.go
```diff
@@ -391,3 +391,33 @@ func (m *Module) Iterator() *BPFObjectIterator {
 		prevMap:  nil,
 	}
 }
+
+func (m *Module) AttachProgs() error {
```
geyslan, line 395: A comment for this method is welcome, explaining the emulation for bpf_object__attach_skeleton

module.go
geyslan, line 412: Ditto about the comment.

selftest/module-attach-detach/main.bpf.c
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    *process = 2021;
+
+    bpf_ringbuf_submit(process, ringbuffer_flags);
+
+    return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
```
geyslan, line 58: nitpicking: add end empty line.

geyslan (changes requested):
@chentao-kernel thank you for this contrib.

As the proposal is to emulate `bpf_object__attach_skeleton` and `bpf_object__detach_skeleton`, some behaviours have to be replicated. I've put comments below.

Please check also for the selftest failure: https://github.com/aquasecurity/libbpfgo/actions/runs/8326584366/job/22782547013?pr=409

module.go
chentao-kernel: I add two apis: AutoLoad and GetProg to solve this.

module.go
chentao-kernel: done

module.go
chentao-kernel: some comments were added.

module.go
chentao-kernel: done

selftest/module-attach-detach/main.bpf.c
chentao-kernel: done

chentao-kernel:
> @chentao-kernel thank you for this contrib.
> 
> As the proposal is to emulate `bpf_object__attach_skeleton` and `bpf_object__detach_skeleton`, some behaviours have to be replicated. I've put comments below.
> 
> Please check also for the selftest failure: https://github.com/aquasecurity/libbpfgo/actions/runs/8326584366/job/22782547013?pr=409

Ok, i will check the failed case.

chentao-kernel:
> > @chentao-kernel thank you for this contrib.
> > As the proposal is to emulate `bpf_object__attach_skeleton` and `bpf_object__detach_skeleton`, some behaviours have to be replicated. I've put comments below.
> > Please check also for the selftest failure: https://github.com/aquasecurity/libbpfgo/actions/runs/8326584366/job/22782547013?pr=409
> 
> Ok, i will check the failed case.

Hi geyslan, the failed case seems confused, it dose not display which case really failed. And i try again, all
case were passed.
image

module.go
```diff
@@ -391,3 +391,53 @@ func (m *Module) Iterator() *BPFObjectIterator {
 		prevMap:  nil,
 	}
 }
+
+func (m *Module) linkExist(prog *BPFProg) bool {
+	for _, link := range m.links {
+		if link.GetProg() == prog {
```
geyslan, line 397: I suspect that this can be misleading or break if the libbpf changes. What guarantees the same pointer coming from it? What do you think of checking against the prog name?

module.go
geyslan: How about not removing while iterating the slice? I know that the lib doesn't have multi-access control, but minimising possible races is better. As our Destroy returns error but DetachProgs should emulate the full destruction of bpf_object__detach_skeleton (that disregards errors), when a err occurs, we should store the non destroyed link with its error (maybe in a local map) and continue iterating.

At the end, assign an empty slice to m.links. If the local map is len > 0, iterate it reassigning the non destroyed links and wrapping/formatting all the errors to be returned.

WDYT?

geyslan:
> Hi geyslan, the failed case seems confused, it dose not display which case really failed. And i try again, all case were passed. 

Ok, disregard this, it was due to [selftest/iter](https://github.com/aquasecurity/libbpfgo/actions/runs/8326584366/job/22782547013?pr=409#step:4:1697), possibly a network issue.

module.go
chentao-kernel: Thanks，your method seems better, i will update it latter.

module.go
chentao-kernel: > I suspect that this can be misleading or break if the libbpf changes. What guarantees the same pointer coming from it? What do you think of checking against the prog name?
> 
> --- EDIT
> 
> I meant this deal with prog returned by NextProgram(), so it wraps again a new BPFProg with a pointer provided from the libbpf. Just being extra cautious here.

you mean like?:
if link.GetProgName() == prog.Name()

geyslan: if possible link.prog.Name(), without another helper.

chentao-kernel: > if possible link.prog.Name(), without another helper.
Got it!

chentao-kernel: > > if possible link.prog.Name(), without another helper.
> > Got it!
done!

module.go
chentao-kernel: > Thanks，your method seems better, i will update it latter.

Done, check it out when you have time, thank you.

geyslan:
@chentao-kernel take a look at this: https://github.com/chentao-kernel/libbpfgo/pull/2

chentao-kernel:
> @chentao-kernel take a look at this: [chentao-kernel#2](https://github.com/chentao-kernel/libbpfgo/pull/2)

Hi, geyslan, thank you for your cooperation, i have merged it.

geyslan (approved):
LGTM

geyslan:
@chentao-kernel thanks again for your contribution.

PR #411, opened 2024-03-27
feat(map): Add GetNextKey and GetValueAndDeleteKey

chentao-kernel:
Add GetNextKey() and GetValueAndDeleteKey() for map and map-low.


chentao-kernel:
> Add GetNextKey() and GetValueAndDeleteKey() for map and map-low.

@geyslan Hi, geyslan Support map apis in todo list, please review when you are free, thanks lot.

map-low.go
```diff
@@ -279,9 +279,17 @@ func (m *BPFMapLow) GetValueFlags(key unsafe.Pointer, flags MapFlag) ([]byte, er
 	return value, nil
 }
 
-// TODO: implement `bpf_map__lookup_and_delete_elem`
-// func (m *BPFMapLow) GetValueAndDeleteKey(key unsafe.Pointer) ([]byte, error) {
-// }
+// GetValueAndDeleteKey gets the value with the given key and delete the key from the map.
+// It returns the value and delete the given key.
+func (m *BPFMapLow) GetValueAndDeleteKey(key unsafe.Pointer) ([]byte, error) {
+	value := make([]byte, m.ValueSize())
```
geyslan, line 285: In `GetValueFlags` above, we took into account calculating the value size via `calcMapValueSize`, even this being only done by the `bpf_map__lookup_elem` version (which in turn calls `bpf_map_lookup_elem_flags`.

I think it's good to have this pre-calculation as a previous step in the upcoming map API as well.

map.go
```diff
@@ -405,9 +405,9 @@ func (m *BPFMap) GetValueFlags(key unsafe.Pointer, flags MapFlag) ([]byte, error
 	return value, nil
 }
 
-// TODO: implement `bpf_map__lookup_and_delete_elem` wrapper
-// func (m *BPFMap) GetValueAndDeleteKey(key unsafe.Pointer) ([]byte, error) {
-// }
+func (m *BPFMap) GetValueAndDeleteKey(key unsafe.Pointer) ([]byte, error) {
+	return m.bpfMapLow.GetValueAndDeleteKey(key)
```
geyslan, line 409: As this has its own correlation in the libbpf API, we should use it instead of relying on the low level API.

`bpf_map__lookup_and_delete_elem`

Use, by instance, `(m *BPFMap) GetValueFlags` structure as a hint.

map.go
```diff
@@ -480,9 +480,9 @@ func (m *BPFMap) DeleteKey(key unsafe.Pointer) error {
 	return nil
 }
 
-// TODO: implement `bpf_map__get_next_key` wrapper
-// func (m *BPFMap) GetNextKey(key unsafe.Pointer) (unsafe.Pointer, error) {
-// }
+func (m *BPFMap) GetNextKey(key unsafe.Pointer) (unsafe.Pointer, error) {
+	return m.bpfMapLow.GetNextKey(key)
```
geyslan, line 484: Ditto: `bpf_map__get_next_key`

geyslan (changes requested):

map-low.go
chentao-kernel: > In `GetValueFlags` above, we took into account calculating the value size via `calcMapValueSize`, even this being only done by the `bpf_map__lookup_elem` version (which in turn calls `bpf_map_lookup_elem_flags`.
> 
> I think it's good to have this pre-calculation as a previous step in the upcoming map API as well.

hi, deyslan, add these apis for look_and_delete, is it ok?
`func (m *BPFMapLow) GetValueAndDeleteKeyFlags(key unsafe.Pointer, flags MapFlag)`
`bpf_map_lookup_and_delete_elem` 

`func (m *BPFMapLow) GetValueAndDeleteKey(key unsafe.Pointer)`
`bpf_map_lookup_and_delete_elem_flags` 

`func (m *BPFMap) GetValueAndDeleteKey(key unsafe.Pointer, flags MapFlag)`
`bpf_map__lookup_and_delete_elem`

geyslan: You can create both, as we already do in GetValue{Flags}, setting the default flag for the shorter version and explaining well in the comments the shorter signature behaviour.

image

Just make sure of wrapping the correct version (for BPFMap and BPFMapLow) as I mentioned [above](https://github.com/aquasecurity/libbpfgo/pull/411/files#r1546773498).

geyslan:
@chentao-kernel I'm baking the next version. Please tell me if you'll have time to finish this so it can be part of it. 👍🏼

chentao-kernel:
> @chentao-kernel I'm baking the next version. Please tell me if you'll have time to finish this so it can be part of it. 👍🏼

Yes, i can finish it

map.go
chentao-kernel: > Ditto: `bpf_map__get_next_key`

done

map.go
chentao-kernel: > As this has its own correlation in the libbpf API, we should use it instead of relying on the low level API.
> 
> `bpf_map__lookup_and_delete_elem`
> 
> Use, by instance, `(m *BPFMap) GetValueFlags` structure as a hint.

done

map-low.go
chentao-kernel: done

chentao-kernel:
> > @chentao-kernel I'm baking the next version. Please tell me if you'll have time to finish this so it can be part of it. 👍🏼
> 
> Yes, i can finish it

@geyslan  update it, please check again

geyslan:
@chentao-kernel I raised this https://github.com/chentao-kernel/libbpfgo/pull/3.

chentao-kernel:
> @chentao-kernel I raised this [chentao-kernel#3](https://github.com/chentao-kernel/libbpfgo/pull/3).

@geyslan hi, geyslan, thank you for your help, it was mergered.

geyslan:
@chentao-kernel, I was rethinking this API and realised that the initial TODO comment led to an erroneous GetNextKey signature as it is not performant (always allocating a buffer to store the fetched key value). This idea was brought up because in the current codebase we have some methods that work as helpers, already instantiating the values and doing the dirty work for the consumer. However, continuing with this approach is limiting the user to a sub-optimized API. So I bring the idea of from this new API we bring 1:1 wrappers and through them we create new helpers (with sugar and without performance guarantee).

I'll close this PR and soon send another one with this idea on top of your work.

In any case, it's a thing that should be addressed in the rest of the codebase, both for new and existing APIs.

chentao-kernel:
> @chentao-kernel, I was rethinking this API and realised that the initial TODO comment led to an erroneous GetNextKey signature as it is not performant (always allocating a buffer to store the fetched key value). This idea was brought up because in the current codebase we have some methods that work as helpers, already instantiating the values and doing the dirty work for the consumer. However, continuing with this approach is limiting the user to a sub-optimized API. So I bring the idea of from this new API we bring 1:1 wrappers and through them we create new helpers (with sugar and without performance guarantee).
> 
> I'll close this PR and soon send another one with this idea on top of your work.
> 
> In any case, it's a thing that should be addressed in the rest of the codebase, both for new and existing APIs.

excellent，l saw your pr.

PR #412, opened 2024-04-01
Optionally enable kernel logging

umanwizard:
Add `NewModuleArgs`'s `KernelLogLevel` arg option.

### DIFF
--- a/module.go
+++ b/module.go
@@ -40,7 +40,7 @@ type NewModuleArgs struct {
 	BPFObjPath      string
 	BPFObjBuff      []byte
 	SkipMemlockBump bool
-	KernelLogLevel  uint32
+	KernelLogLevel  C.uint
 }
 
 func NewModuleFromFile(bpfObjPath string) (*Module, error) {
--- a/module.go
+++ b/module.go
@@ -40,7 +40,7 @@ type NewModuleArgs struct {
 	BPFObjPath      string
 	BPFObjBuff      []byte
 	SkipMemlockBump bool
-	KernelLogLevel  C.uint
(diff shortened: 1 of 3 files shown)


module.go
```diff
@@ -40,6 +40,7 @@ type NewModuleArgs struct {
 	BPFObjPath      string
 	BPFObjBuff      []byte
 	SkipMemlockBump bool
+	KernelLogLevel  C.uint
```
geyslan, line 43:
```suggestion
	KernelLogLevel  uint32
```

Use go types here and do the conversion inside the wrappers.

module.go
```diff
@@ -138,6 +139,7 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 	if optsC == nil {
 		return nil, fmt.Errorf("failed to create bpf_object_open_opts: %w", errno)
 	}
+	optsC.kernel_log_level = args.KernelLogLevel
```
geyslan, line 142: Instantiate the C object `kernelLogLevelC` right below `bpfBuffSizeC := C.size_t(len(args.BPFObjBuff))`.

That's to avoid direct assignments to C structs since cgo doesn't work 100% of the cases. With that in mind, you should change the `cgo_bpf_object_open_opts_new()` helper to receive `kernelLogLevelC` and assign it properly.

geyslan (changes requested):
@umanwizard thanks for your contribution.

I've put a couple of comments. Let me know.

umanwizard:
thanks, updated.

geyslan:
@umanwizard thanks again for your contrib.

Issue #429, opened 2024-05-14
Static builds with newer libelf require `-l zstd`


bobrik:
Ubuntu 24.04 links `libelf.so` against `libzstd.so`. This doesn't matter at all for dynamically linked builds, but if you want a link a fully static binary, you need `-l zstd` in `CGO_LDFLAGS`. Currently `libbpfgo` only has:

```
#cgo LDFLAGS: -lelf -lz
```

I worked around this in `ebpf_exporter`:

* https://github.com/cloudflare/ebpf_exporter/blob/b908d3ba7/Makefile#L34-L37

```makefile
# If libelf is new and it links to zstd, then zstd needs to be manually linked in.
ifneq ($(shell ldd $(shell /sbin/ldconfig -n -p | grep libelf.so.1 | awk '{ print $$NF }') | grep zstd),)
CGO_LDFLAGS := $(CGO_LDFLAGS) -l zstd
endif
```

Perhaps there's a better way in `libbpfgo` itself.

geyslan:
@bobrik thanks for reporting this.

@trvll is it the same that you facing? Could you test it?

bobrik:
I think you can replicate this in Github actions by bumping Ubuntu 22.04 to 24.04.

geyslan:
Thanks @bobrik. zstd is indeed a requirement depending on the environment when a static build is the target. I'll be bumping this repo as soon as possible and tackling that.

PR #431, opened 2024-05-22
Bumps

geyslan:
Close: #429

commit 517e2d567876d140d5ddc30aabb96283bbc9b717

    chore: code format as clang-format-14

commit 46560d5a49dfb383543632e247b93da8aad37c37

    chore: bump ubuntu images, go, clang, actions...
    
    clang-format pinned to the same version of the clang compiler.
    
    Co-authored-by: Raphael Campos <raphaelcampos.rp@gmail.com>

commit 7b0db19144822ec5e83d72dca50cdfa25587199d

    chore: add libzstd-dev to preparation

commit 098f20b0a6641b7f5a0d4eab4157c0189961d08a

    fix(build): set required ldflags
    
    For static build, get pc info from the freshly built libbpf.a instead of
    the system one.
    For dynamic build, set only the required ldflags for the system libbpf.

commit 76bd3af1bb691f5c78d5ee3321aaca3c49eccfca

    chore: bump go.mod files
    
    go version to 1.22.0
    go toolchain to 1.22.3


.github/workflows/pr.yaml
```diff
@@ -54,13 +54,13 @@ jobs:
         shell: bash
   libbpfgo-unit-tests:
     name: libbpfgo Unit Tests
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
-        go-version: [ 1.18, 1.19, '1.20', '1.21', 'stable' ]
+        go-version: [ '1.21', '1.22', 'stable' ]
```
geyslan, line 60: @rscampos as we're using toolchain now, it's available from 1.21, that's why I removed the older ones.

geyslan:
@rscampos it would be nice to test it in arm64.

builder/prepare-ubuntu.sh
```diff
@@ -7,9 +7,21 @@
...
+    x86_64|amd64)
+        ARCH="amd64"
+        ;;
+    aarch64|arm64)
+        ARCH="arm64"
+        ;;
+    *)
+        die "unsupported architecture ${ARCH}"
+        ;;
+esac
```
geyslan, line 24: ARCH is not currently used but might be if we need to download binaries instead of installing them via the package system.

geyslan:
@rscampos as you mentioned in https://github.com/geyslan/libbpfgo/pull/8#issuecomment-2130237649, next step is to decide what style .clang-format will enforce to us. 👍🏼

rscampos:
@geyslan I did some tests on arm... just need to change to clang-format-14, all good :)

rscampos (approved):
LGTM

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/aquasecurity/libbpfgo
 
-go 1.21
+go 1.22.0
```
bobrik, line 3: Is it necessary to bump the minimum Go version?

This has a knockdown effect on everything that depends on `libbpfgo`:

```diff
$ git diff go.mod
diff --git a/go.mod b/go.mod
index dd5b16f..f1fece8 100644
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,12 @@
 module github.com/cloudflare/ebpf_exporter/v2

-go 1.21
+go 1.22.0
+
+toolchain go1.22.3

 require (
        github.com/alecthomas/kingpin/v2 v2.4.0
-       github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4
+       github.com/aquasecurity/libbpfgo v0.7.0-libbpf-1.4.0.20240524222422-daeb5d0059fe
        github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
        github.com/elastic/go-perf v0.0.0-20191212140718-9c656876f595
        github.com/iovisor/gobpf v0.2.0
```

I think it also forces Go upgrade from v1.22.2 to v1.22.3:

```
go: downloading go1.22.3 (linux/arm64)
```

geyslan: > Is it necessary to bump the minimum Go version?
> 
> This has a knockdown effect on everything that depends on `libbpfgo`:

Yes, it really enforces consumers to align versions.

There are improvements in 1.22 that can be useful for all of us, including performance ones: https://tip.golang.org/doc/go1.22

@bobrik can't you bump it at your side?

> 
> I think it also forces Go upgrade from v1.22.2 to v1.22.3:
> 
> ```
> go: downloading go1.22.3 (linux/arm64)
> ```

Don't worry about this, it's the toolchain requirement, it won't update your system go. It downloads and uses for the current project.

bobrik: You are not supposed to dictate what go version is necessary for the consumers unless it's actually necessary:

* https://go.dev/doc/modules/gomod-ref#go

No matter how good the latest Go version is, if it's not a requirement, you are not supposed to force it. You set the minimum supported Go version instead. Go supports two major releases at any time and it seems reasonable to do at least the same thing.

People with patched go versions might not have an ability to upgrade to the latest release.

I don't think there's a good reason to set `toolchain` for a library ever.

geyslan: You're absolutely right. I'm going to revert (soon) this PR by splitting the fix and other chores into their own PRs. 👍🏼

bobrik: Thanks, I appreciate it.

PR #433, opened 2024-05-28
Revert "Bumps"

geyslan:
Reverts aquasecurity/libbpfgo#431

As well [reminded](https://github.com/aquasecurity/libbpfgo/pull/431#discussion_r1616431983) by @bobrik, even if we can, we shouldn't force consumers to use specific go versions by declaring toolchain or raising the minimum version.

As a good practice, whenever possible, we should keep libbpfgo aligned with the last two go major releases.

This reverted PR will be reworked in upcoming ones.


rscampos (approved):
LGTM

PR #434, opened 2024-05-28
Fix `CGO_LDFLAGS*`

geyslan:
Close: https://github.com/aquasecurity/libbpfgo/issues/429

commit 63c7b910a7b003e0044a4749dc22322962218ec2

    chore: add libzstd-dev to preparation
    
    libzstd may be a dependency on libelf, which is a dependency on libbpf.

commit 85ef7598d64064d5100d965f09eb1bf6d2717c01

    fix(build): set required ldflags
    
    For static build, get pc info from the freshly built libbpf.a instead of
    the system one.
    For dynamic build, set only the required ldflags for the system libbpf.


rscampos (approved):
LGTM

.github/actions/build-dependencies/action.yaml
```diff
@@ -20,7 +20,7 @@ runs:
         sudo apt-get install --yes pkgconf
         sudo apt-get install --yes llvm-12 clang-12
         sudo apt-get install --yes clang-format-12
-        sudo apt-get install --yes zlib1g-dev libelf-dev
+        sudo apt-get install --yes libelf-dev libzstd-dev zlib1g-dev
```
bobrik, line 23: I think you only really need `libelf-dev`, the rest is automatically pulled in:

* https://packages.ubuntu.com/noble/libelf-dev

geyslan: It's indeed. In fact, it could just be the main `libbpf-dev`, but it might be good to have them listed so that users from different environments (even tarball ones) can take a look at what is needed. @rscampos FYI.

Issue #340, opened 2023-07-20
NewModuleFromBufferArgs should not specify BTF if there isn't one


MartinDai:
like #81

PR #436, opened 2024-05-29
libbpfgo: do not set btfFilePathC when no BTF is given in ModuleFromBuffer

MarcWort:
like 5034da49eaac59789820fb82de1191ff1b91ede7
Fixes: #340

### DIFF
--- a/module.go
+++ b/module.go
@@ -119,14 +119,21 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 		return nil, err
 	}
 
-	if args.BTFObjPath == "" {
-		args.BTFObjPath = "/sys/kernel/btf/vmlinux"
+	var btfFilePathC *C.char
+	var kConfigPathC *C.char
+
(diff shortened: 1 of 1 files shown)


module.go
```diff
@@ -119,14 +119,21 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 		return nil, err
 	}
 
-	if args.BTFObjPath == "" {
-		args.BTFObjPath = "/sys/kernel/btf/vmlinux"
```
geyslan, line 123: I suppose we should keep the hard-coded path when not set by the caller.

MarcWort: It should have the same behavior as `NewModuleFromFileArgs` ( see https://github.com/aquasecurity/libbpfgo/blob/main/module.go#L71).

It is currently not possible to set no BTFObjPath.

geyslan (approved):
LGTM

@MarcWort thanks for contributing!

PR #437, opened 2024-05-29
feat(ksymbols): reintroduce lazy symbol queries

NDStrahilevitz:
Add new required symbol and addresses lists to the KSymbolTable helper. When either list is given, the data structure will only save required symbols during scanning. This should reduce memory usage for users who can mostly guarentee that symbols are known ahead of time.

Dynamic querying is still allowed in this mode, though it will incur a penalty as rescanning is required.

### DIFF
--- a/helpers/kernel_symbols.go
+++ b/helpers/kernel_symbols.go
@@ -29,13 +29,9 @@ func symNotFoundErr(v interface{}) error {
 // Interface implementation
 //
 
-type name struct {
-	name string
-}
+type name string
 
-type addr struct {
-	addr uint64
-}
+type addr uint64
 
 type nameAndOwner struct {
 	name  string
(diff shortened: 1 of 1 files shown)


geyslan:
Going on this soon.

helpers/kernel_symbols.go
```diff
@@ -48,41 +44,79 @@ type addrAndOwner struct {
...
 //
 // Getters (return a copy of the symbol for thread safety).
 //
 
 // TextSegmentContains returns true if the given address is in the kernel text segment.
 func (k *KernelSymbolTable) TextSegmentContains(addr uint64) (bool, error) {
-	k.updateLock.RLock()
-	defer k.updateLock.RUnlock()
+	k.updateLock.Lock()
+	defer k.updateLock.Unlock()
```
geyslan, line 97: As the goal is to squeeze as much as possible (optimise wise), what do you think unlocking right after the getTextSegmentAddresses() call?

helpers/kernel_symbols.go
```diff
@@ -225,6 +308,13 @@ func (k *KernelSymbolTable) processLines(chans []chan *KernelSymbol) error {
 			continue
 		}
 		if sym := parseLine(fields); sym != nil {
+			if k.onlyRequired {
```
geyslan, line 311: Can't we assume at this stage that if `!k.onlyRequired` we should continue?

helpers/kernel_symbols.go
```diff
@@ -29,13 +29,9 @@ func symNotFoundErr(v interface{}) error {
 // Interface implementation
 //
 
-type name struct {
-	name string
-}
+type name string
```
geyslan, line 32: Good call.

helpers/kernel_symbols.go
```diff
@@ -48,41 +44,79 @@ type addrAndOwner struct {
...
+// NOTE: If table is in required only mode, method will add the new symbol
+// and rescan the file.
 func (k *KernelSymbolTable) GetSymbolByName(n string) ([]KernelSymbol, error) {
-	k.updateLock.RLock()
-	defer k.updateLock.RUnlock()
+	k.updateLock.Lock()
+	defer k.updateLock.Unlock()
 
-	symbols, exist := k.symbols[name{n}]
+	err := k.validateOrAddRequiredSym(name(n))
```
geyslan, line 114: What about returning `symbols` and `exist` from `validateOrAddRequiredSym`, so we could release the lock right after it returns checking them later.

geyslan (approved):
LGTM, put some questions.

geyslan:
Closing this in favour of https://github.com/aquasecurity/tracee/pull/4095

PR #438, opened 2024-05-30
chore(prog): Add attach/detach legacy api for xdp

chentao-kernel:
The AttachXDP api use bpf link to attach xdp prog, AttachXDPLegacy/DetachXDPLegacy apis corresponds to bpf_xdp_attach/bpf_xdp_detach, which can be used on old kernel, so add it.

### DIFF
--- a/prog-common.go
+++ b/prog-common.go
@@ -237,3 +237,19 @@ const (
 	BPFFAllowMulti    AttachFlag = C.BPF_F_ALLOW_MULTI
 	BPFFReplace       AttachFlag = C.BPF_F_REPLACE
 )
+
+//
+// XDPFlags
+//
+
+type XDPFlags uint32
+
+const (
+	XDPFlagsUpdateIfNoExist XDPFlags = C.XDP_FLAGS_UPDATE_IF_NOEXIST
+	XDPFlagsSkbMode         XDPFlags = C.XDP_FLAGS_SKB_MODE
+	XDPFlagsDrvMode         XDPFlags = C.XDP_FLAGS_DRV_MODE
+	XDPFlagsHwMode          XDPFlags = C.XDP_FLAGS_HW_MODE
+	XDPFlagsReplace         XDPFlags = C.XDP_FLAGS_REPLACE
+	XDPFlagsModes           XDPFlags = C.XDP_FLAGS_MODES
+	XDPFlagsMask            XDPFlags = C.XDP_FLAGS_MASK
+)
--- a/prog.go
+++ b/prog.go
@@ -308,6 +308,44 @@ func (p *BPFProg) AttachXDP(deviceName string) (*BPFLink, error) {
 	return bpfLink, nil
 }
 
+func (p *BPFProg) AttachXDPLegacy(deviceName string, flag XDPFlags) error {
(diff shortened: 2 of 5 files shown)


prog.go
```diff
@@ -308,6 +308,45 @@ func (p *BPFProg) AttachXDP(deviceName string) (*BPFLink, error) {
 	return bpfLink, nil
 }
 
+func (p *BPFProg) AttachXDPLegacy(deviceName string, flag XDPFlags) error {
+	optsC := &C.struct_bpf_xdp_attach_opts{
+		sz:          C.ulong(unsafe.Sizeof(C.struct_bpf_xdp_attach_opts{})),
+		old_prog_fd: 0,
+	}
```
geyslan, line 315: Could you convert these instantiations to use the `structure handlers` pattern found in `libbpfgo.h`, since cgo has problems with structure packaging and field offsets? By instance:

- cgo_bpf_xdp_attach_opts_new
- cgo_bpf_xdp_attach_opts_free

prog.go
```diff
@@ -308,6 +308,45 @@ func (p *BPFProg) AttachXDP(deviceName string) (*BPFLink, error) {
...
+
+	return nil
+}
+
+func (p *BPFProg) DetachXDPLegacy(deviceName string, flag XDPFlags) error {
+	optsC := &C.struct_bpf_xdp_attach_opts{
+		sz:          C.ulong(unsafe.Sizeof(C.struct_bpf_xdp_attach_opts{})),
+		old_prog_fd: C.int(p.FileDescriptor()),
+	}
```
geyslan, line 335: ditto

prog-common.go
```diff
@@ -48,6 +48,19 @@ const (
...
+const (
+	XDPFlagsUpdateIfNoExist XDPFlags = C.XDP_FLAGS_UPDATE_IF_NOEXIST
+	XDPFlagsSkbMode         XDPFlags = C.XDP_FLAGS_SKB_MODE
+	XDPFlagsDrvMode         XDPFlags = C.XDP_FLAGS_DRV_MODE
+	XDPFlagsHwMode          XDPFlags = C.XDP_FLAGS_HW_MODE
+	XDPFlagsReplace         XDPFlags = C.XDP_FLAGS_REPLACE
+	XDPFlagsModes           XDPFlags = C.XDP_FLAGS_MODES
+	XDPFlagsMask            XDPFlags = C.XDP_FLAGS_MASK
+)
+
```
geyslan, line 63: Please move this block to the end of the file following the comment pattern.

geyslan (changes requested):
That's great. I placed some requests. As soon as they are addressed I will do the final tests. Tks.

chentao-kernel:
> That's great. I placed some requests. As soon as they are addressed I will do the final tests. Tks.

Hi, geyslan, thank you for your reply, i will do it later!

prog.go
chentao-kernel: done

prog.go
chentao-kernel: done

prog-common.go
chentao-kernel: done

geyslan (approved):
LGTM. @chentao-kernel thanks a lot.

PR #428, opened 2024-05-11
Add method Run to BPFProg(BPF_PROG_RUN)

sc07kvm:
The written code passes a **bpf_prog_test_run_opts** from libbpf. This is a good feature for writing tests on eBPF code. 
References:

- https://docs.kernel.org/bpf/bpf_prog_run.html
- https://docs.kernel.org/userspace-api/ebpf/syscall.html

Example of a testing library in Cilium built on this feature:

- https://docs.cilium.io/en/stable/contributing/testing/bpf/

PR #440, opened 2024-06-04
feat: Add method Run to BPFProg

geyslan:
feat: Add method Run to BPFProg
    
    Run method is used to load and run the BPF program without attaching it
    to any hook. This is useful for testing the BPF program.
    
    It makes use of the bpf(BPF_PROG_TEST_RUN) via bpf_prog_test_run_opts().
    
    The prog-run selftest is added to test the Run method. More examples
    can be found in the linux kernel source code:
    tools/testing/selftests/bpf/prog_tests.
    
    Original code #428 was authored by sc07kvm, and this commiter have only
    made cosmetics adjustments.

### DIFF
--- a/prog.go
+++ b/prog.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
+	"time"
 	"unsafe"
 )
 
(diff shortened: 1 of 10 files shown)


prog.go
```diff
@@ -629,3 +630,160 @@ func (p *BPFProg) DetachGenericFD(targetFd int, attachType BPFAttachType) error
...
+//	        DataOut: make([]byte, 0, 14),
+//	        DataSizeOut: 14,
+//	        Repeat: 1,
+//	    }
+//	    prog.Run(&opts)
+//	    if opts.RetVal != 1 {
+//	        t.Errorf("result = %d; want 1", opts.RetVal)
+//	    }
+//	}
+func (p *BPFProg) Run(opts *RunOpts) error {
```
NDStrahilevitz, line 773: I think it would be nice to add an error here when the program type isn't compatible (rather than relying on an internal failure).

NDStrahilevitz (approved):
LGTM, a nice addition indeed.

prog.go
geyslan: It will be dependent on kernel version support, e.g.:

- https://elixir.bootlin.com/linux/v6.9.3/source/net/bpf/test_run.c#L728 (supported)
- https://elixir.bootlin.com/linux/v6.9.3/source/include/linux/bpf.h#L3075 (unsupported)
- https://elixir.bootlin.com/linux/v6.9.3/source/include/linux/bpf.h#L2817 (unsupported)

The output for `ENOTSUPP` is `failed to run program: errno 524`. https://elixir.bootlin.com/linux/v6.9.3/source/include/linux/errno.h#L27

`.Error()` doesn't find a correct entry, since "not supported" for syscall is `ENOTSUP` 95. https://github.com/golang/sys/blob/master/unix/zerrors_linux_amd64.go#L614

geyslan: WDYT? If you agree, we must check all others returns for this `ENOTSUPP`.

```diff
diff --git a/prog.go b/prog.go
index e47c8ec..56ea57e 100644
--- a/prog.go
+++ b/prog.go
@@ -779,7 +779,11 @@ func (p *BPFProg) Run(opts *RunOpts) error {
 
        retC := C.bpf_prog_test_run_opts(C.int(p.FileDescriptor()), optsC)
        if retC < 0 {
-               return fmt.Errorf("failed to run program: %w", syscall.Errno(-retC))
+               errno := syscall.Errno(-retC)
+               if errno == 524 { // ENOTSUPP https://elixir.bootlin.com/linux/v6.9.3/source/include/linux/errno.h#L27
+                       errno = syscall.ENOTSUP
+               }
+               return fmt.Errorf("failed to run program: %s", errno)
        }
 
        // update runOpts with the values from the kernel and libbpf
```

geyslan: Anyway, merging. If decided for it, we can tackle it in a bunch.

PR #450, opened 2024-07-12
fix(build): include libbpf during libbpfgo compilation

rscampos:
An issue appears when we try to compile libbpfgo on an older kernel that doesn't have the XDP_FLAGS_REPLACE defined in
/usr/include/linux/if_link.h. To overcome this, it is necessary to include the directory of libbpf under 3rdparty during the compilation, since install_uapi_headers doesn't install if_link.h.


geyslan (approved):
LGTM

PR #451, opened 2024-07-26
fix: set 'elf' to nil after closing in BPFLoadObject for GC

rscampos:
In the BPFLoadObject function, the 'elf' field is closed, but it is not set to nil. Setting it to nil is necessary for it to be freed by the garbage collector.

### DIFF
--- a/module.go
+++ b/module.go
@@ -202,6 +202,7 @@ func (m *Module) BPFLoadObject() error {
 	}
 	m.loaded = true
 	m.elf.Close()
+	m.elf = nil
 
 	return nil
 }


rscampos:
In order to test this commit, use the following snip before and after `BPFLoadObject()`:
```
var memBefore runtime.MemStats
runtime.ReadMemStats(&memBefore)
fmt.Println("Alloc mem before: ", memBefore.Alloc)
```

**Some results:**
Without this commit:
```
sudo ./run.sh main-static
Memory stats before:
Alloc = 85952 bytes
Memory stats after:
Alloc = 85984 bytes
```

With this commit:
```
sudo ./run.sh main-static
Memory stats before:
Alloc = 85952 bytes
Memory stats after:
Alloc = 80144 bytes
```

geyslan (approved):
Good catch. ❤️ It seems safe to nullify it since `m.loaded` is checked first in the related logics.

PR #473, opened 2025-01-09
chore!: remove helpers package

geyslan:
a331a714f01d2f20b78f59e7bf624cab2ddb3bcc chore!: remove helpers package
    
    Conclusion of #458


.github/dependabot.yml
```diff
@@ -4,15 +4,9 @@ updates:
...
       interval: daily
-  # - package-ecosystem: "github-actions"
-  #   directory: "/helpers"
-  #   schedule:
-  #     interval: daily
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+      - "/"
+      - "selftest/**/*"
```
geyslan, line 10: Opportunistic change.

.github/dependabot.yml
```diff
@@ -4,15 +4,9 @@ updates:
...
       interval: daily
-  # - package-ecosystem: "github-actions"
-  #   directory: "/helpers"
-  #   schedule:
-  #     interval: daily
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+      - "/"
+      - "/selftest/**/*"
```
geyslan, line 10: Opportunistic change.

geyslan:
@oshaked1 @OriGlassman please signal us when your end is ready for removal.

selftest/tracing-by-offset/main.go
```diff
@@ -9,14 +9,14 @@ import (
...
-	"github.com/aquasecurity/libbpfgo/helpers"
 	"github.com/aquasecurity/libbpfgo/selftest/common"
+	"github.com/aquasecurity/tracee/pkg/symbols"
 )
 
 func main() {
 	funcName := fmt.Sprintf("__%s_sys_mmap", common.KSymArch())
 
-	kst, err := helpers.NewKernelSymbolTable()
+	kst, err := symbols.NewKernelSymbolTable(false, false)
```
geyslan, line 19: @yanivagman as you can see, as symbols pkg is not under common, it demands a full import of Tracee, changing the [go version and toolchain](https://github.com/aquasecurity/libbpfgo/pull/473/files#diff-49b92141949f2a52b8dd64f7e611f239a7145f3b0cf165b731f8797050d81fe1R3-R5). This is not a problem for these two selftests. JFYI.

geyslan: You're right, we can get rid of that dependency completely.

PR #475, opened 2025-01-16
fix: probe bpf func

rscampos:
- Improved BPFHelperIsSupported by enhancing error reporting to include detailed messages for unsupported helpers and unexpected errno values returned by libbpf;
- Added logic to handle unexpected errno states and clarified support assumptions based on retC == 1 when running with capabilities.

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -1,8 +1,8 @@
 package libbpfgo
 
 import (
-	"errors"
 	"fmt"
+	"strings"
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


geyslan (dismissed):
LGTM

geyslan:
As we discussed offline, it's better to return to origins and make the libbpfgo, mostly for this matter, return ret and errno as it is. Please provide comments about the findings related to libbpf and kernel. Tks!

geyslan (approved):

PR #479, opened 2025-02-12
Fix incorrect prctl values and add new ones

oshaked1:


geyslan:
@oshaked1 helpers are deprecated and marked to be removed soon, see: https://github.com/aquasecurity/libbpfgo/pull/473

This must be handled in Tracee's https://github.com/aquasecurity/tracee/pull/4090.

helpers/argumentParsers.go
```diff
@@ -579,13 +581,39 @@ const (
...
+	PR_SET_SYSCALL_USER_DISPATCH
+	PR_PAC_SET_ENABLED_KEYS
+	PR_PAC_GET_ENABLED_KEYS
+	PR_SCHED_CORE
+	PR_SME_SET_VL
+	PR_SME_GET_VL
+	PR_SET_MDWE
+	PR_GET_MDWE
+	PR_SET_MEMORY_MERGE
+	PR_GET_MEMORY_MERGE
```
geyslan, line 604: Some of those were already inserted in Tracee's: https://github.com/aquasecurity/tracee/blob/1150c947b726ba22f0432106d85abc389ffb8699/pkg/events/parsers/data_parsers.go#L496-L502

Other were ignored due to from unsupported archs.

geyslan: Ok, we've discussed offline, merging it for now.

geyslan (approved):

PR #493, opened 2025-07-01
cgo polling loop

NDStrahilevitz:
feat(buffer): move polling to cgo loop
    
Polling loops are cgo hotpaths in consuming applications. Since cgo calls
are rather expensive, moving the polling loop to cgo should significantly
reduce the number of cgo calls.

Now the go part only handles orchestration of the internal c loop.
An atomic flag is used to signal the c loop to exit.

### DIFF
--- a/buf-perf.go
+++ b/buf-perf.go
@@ -9,7 +9,9 @@ import "C"
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"syscall"
+	"unsafe"
 )
 
 //
@@ -22,15 +24,14 @@ type PerfBuffer struct {
 	slot       uint
 	eventsChan chan []byte
 	lostChan   chan uint64
-	stop       chan struct{}
 	closed     bool
 	wg         sync.WaitGroup
+	stopFlag   uint32 // use with atomic operations
 }
 
 // Poll will wait until timeout in milliseconds to gather
 // data from the perf buffer.
 func (pb *PerfBuffer) Poll(timeout int) {
-	pb.stop = make(chan struct{})
 	pb.wg.Add(1)
 	go pb.poll(timeout)
 }
@@ -41,12 +42,12 @@ func (pb *PerfBuffer) Start() {
 }
 
 func (pb *PerfBuffer) Stop() {
-	if pb.stop == nil {
+	if atomic.LoadUint32(&pb.stopFlag) == 1 {
 		return
 	}
 
 	// Signal the poll goroutine to exit
-	close(pb.stop)
+	atomic.StoreUint32(&pb.stopFlag, 1)
 
 	// The event and lost channels should be drained here since the consumer
 	// may have stopped at this point. Failure to drain it will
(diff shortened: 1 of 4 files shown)


module.go
```diff
@@ -436,8 +442,14 @@ func (m *Module) InitPerfBuf(mapName string, eventsChan chan []byte, lostChan ch
 		return nil, fmt.Errorf("failed to initialize perf buffer: %w", errno)
 	}
 
+	stopFlag := C.cgo_setup_buffer_stop_flag()
+	if stopFlag == nil {
```
geyslan, line 446: ditto.

buf-perf.go
```diff
@@ -93,20 +92,29 @@ func (pb *PerfBuffer) Close() {
 func (pb *PerfBuffer) poll(timeout int) error {
 	defer pb.wg.Done()
 
+	const maxRetries = 3
```
geyslan, line 95: I believe that maxRetries should be a public option.

geyslan (approved):
LGTM overall. Just put some doubts.

module.go
```diff
@@ -376,10 +376,16 @@ func (m *Module) InitRingBuf(mapName string, eventsChan chan []byte) (*RingBuffe
 		return nil, fmt.Errorf("failed to initialize ring buffer: %w", errno)
 	}
 
+	stopFlag := C.cgo_setup_buffer_stop_flag()
+	if stopFlag == nil {
```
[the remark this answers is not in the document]
NDStrahilevitz: playing with this kind of code in the last week, atomic syntax seems to work perfectly well as long as you make the right conversions. Do you mean we should have a c function doing the `==NULL` check?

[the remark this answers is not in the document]
geyslan: > Do you mean we should have a c function doing the ==NULL check?

No worries, I get that cgo_setup_buffer_stop_flag only creates the buffer.

buf-perf.go
```diff
@@ -40,13 +41,17 @@ func (pb *PerfBuffer) Start() {
 	pb.Poll(300)
 }
 
+func (pb *PerfBuffer) stopped() bool {
+	return pb.stopFlag.Load() == 1
+}
+
 func (pb *PerfBuffer) Stop() {
-	if pb.stop == nil {
+	if pb.stopped() {
```
yanivagman, line 49: nit: since stopped function is not exported, I think it will be clearer inline

yanivagman (approved):
LGTM

buf-perf.go
```diff
@@ -22,15 +24,14 @@ type PerfBuffer struct {
 	slot       uint
 	eventsChan chan []byte
 	lostChan   chan uint64
-	stop       chan struct{}
 	closed     bool
 	wg         sync.WaitGroup
+	stopFlag   atomic.Uint32 // atomic
```
yanivagman, line 29: unit tests seem to fail on this.
You can define it as
`stopFlag   int32`
Then use
`atomic.StoreInt32(&pb.stopFlag, 1)`

It will also make the cast in the poll function more straight forward since it is an int and not atomic type:
(*C.int)(unsafe.Pointer(&pb.stopFlag))

geyslan: Unit tests fail for Go 1.18 since it was added in 1.19. So, we could deprecate 1.18.

PR #501, opened 2025-08-26
fix(libbpfgo): fix support checks false case error

AlonZivony:
Fix a bug where the support functions would return false with a error for the case the support check return false without errors.


geyslan (approved):
LGTM

PR #502, opened 2025-08-26
selftest/iter: improve test reliability and debugging

geyslan:
f08f80d193b3db54a04d59e8a623cc100e625dcc

    selftest/iter: improve test reliability and debugging
    
    Continuation of 68365063.
    
    - Add BPF-side filtering to only process "sleep" tasks, reducing
      iterator output volume and improving test performance.
    - Increase sleep duration from 2s to 5s to allow processes to register
      in the kernel task list.
    - Add 1s delay after iterator attachment to ensure it's ready before
      starting test processes.
    - Add comprehensive logging to track process creation, iterator output,
      and matching results for easier troubleshooting.
    - Include parent PID in process start logs for better visibility.
    
    These changes address race conditions that could cause the test to fail
    intermittently when processes weren't fully registered before the
    iterator read them.


trvll (approved):
LGTM

selftest/iter/main.bpf.c
```diff
@@ -4,6 +4,9 @@
 
 #include <bpf/bpf_helpers.h>
 
+const char sleep_comm[] = "sleep";
+const int sleep_comm_size = sizeof(sleep_comm);
```
yanivagman, line 8: nit: this sizeof can be in the function itself since it is calculated in compile time

PR #505, opened 2025-08-28
feat(maps): add error for no inner map

AlonZivony:
Use a defined error for calling InnerMapInfo for a map which has no inner map. This allow the caller to distinguish between error in logic to a call on an unsupported map.

### DIFF
--- a/map.go
+++ b/map.go
@@ -7,6 +7,7 @@ package libbpfgo
 import "C"
 
 import (
+	"errors"
 	"fmt"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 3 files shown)


geyslan:
@AlonZivony, would you mind updating [selftest/map-innerinfo/main.go](https://github.com/aquasecurity/libbpfgo/blob/38a101887886b2e15c234f790b366d1ee8f63ef4/selftest/map-innerinfo/main.go) to check the new error?

yanivagman (approved):
LGTM

Issue #489, opened 2025-04-30
[Question] Do we still need bumpMemlockRlimit?


maxgio92:
Hi,
I've noticed that the load

geyslan:
> My concern is to reduce privilege requirements for "almost" unprivileged execution of programs using libbpfgo.

You have a good point. Adding context, #265 reintroduced `bumpMemlockRlimit()`.

> Overall, if I am not wrong, this is potentially the only remaining case where we need setrlimit which requires to have the SYS_RESOURCE capability.

That's a bummer. Do you have any suggestions on handling this remaining corner case?

maxgio92:
Hi @geyslan, I'm really sorry about my very late response.
Thank you for your rapid feedback, really appreciate it.

I'm putting my head back to it and be back asap.

Edit: theoretically `SkipMemlockBump` introduced [here](https://github.com/aquasecurity/libbpfgo/pull/337) should allow to not require the `sys_resource` capability.

maxgio92:
Oh, I see! The patch to [Add SkipMemlockBump to NewModuleArgs](https://github.com/aquasecurity/libbpfgo/pull/337) has only been applied to [NewModuleFromFileArgs](https://github.com/aquasecurity/libbpfgo/blob/7b1853515629447a9204a9bf1db5f43edc743a24/module.go#L61) and not to [NewModuleFromBufferArgs](https://github.com/aquasecurity/libbpfgo/blob/7b1853515629447a9204a9bf1db5f43edc743a24/module.go#L119)

maxgio92:
@geyslan I took a look about the libbpf support for cg mem accounting and its probe to support kernels <5.11, with fallback to rlimit memlocks, and it seems I was wrong:

https://github.com/torvalds/linux/blob/0739473694c4878513031006829f1030ec850bc2/net/core/filter.c#L8044

the bpf_ktime_get_coarse_ns used in the libbpf probe for memcg support, seems supported by the socket filter prog type that the probe function loads. I probably have checked [docs.ebpf.io](https://docs.ebpf.io/linux/helper-function/bpf_ktime_get_coarse_ns), which seems not up to date.

Nevertheless, thanks for the rapid review. I'll bring in the just merged patch into [xcover](github.com/maxgio92/xcover), in order to try to reduce further the required capabilities.

PR #513, opened 2025-10-10
Add SkipMemlockBump to NewModuleBufferArgs

maxgio92:
This is part of the patch [1] that was applied to NewModuleFromFileArgs, but not to NewModuleFromBufferArgs.
This commit applies the same patch to NewModuleFromBufferArgs.

The goal was to remove the `CAP_SYS_RESOURCE` requirement since the rlimit bump was the only remaining one to require that.
Skipping the memory lock bump with the BPF module option introduced by the initial patch [1], will make the program to rely on libbpf that relies on cgroup managed memory accounting [2].

1. https://github.com/aquasecurity/libbpfgo/commit/d2cc6f8a97da4fabbf1613dea7a5a9a1656463bd
2. https://lore.kernel.org/bpf/20201201215900.3569844-1-guro@fb.com/t/

See also #489

### DIFF
--- a/module.go
+++ b/module.go
@@ -115,9 +115,12 @@ func NewModuleFromBufferArgs(args NewModuleArgs) (*Module, error) {
 	}
 	C.cgo_libbpf_set_print_fn()
 
-	// TODO: remove this once libbpf memory limit bump issue is solved
-	if err := bumpMemlockRlimit(); err != nil {
-		return nil, err
+	// If skipped, we rely on libbpf to do the bumping if deemed necessary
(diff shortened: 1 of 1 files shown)


geyslan:
I believe this solve the puzzle (missing part) of NewModuleBufferArgs. Tks @maxgio92. LMK when ready.

maxgio92:
Thanks @geyslan. CI seems happy, marking it as ready for review :)

geyslan (approved):

PR #523, opened 2026-03-24
feat: add support for bpf_program__attach_uprobe_opts

### LINKED ISSUE #522: [Proposal] Support bpf_cookie in single-uprobe attachment
### Background

libbpf exposes `bpf_program__attach_uprobe_opts` which accepts a `struct bpf_uprobe_opts` containing a `bpf_cookie` field. This allows attaching a BPF program to a single uprobe (via the classic `perf_event_open` path) while associating a per-attachment cookie that the program can retrieve at runtime via `bpf_get_attach_cookie()`.

libbpfgo currently wraps the no-cookie variant (`bpf_program__attach_uprobe`) but does not expose the opts variant.

### Problem

There is no way in libbpfgo to attach a single uprobe with a per-attachment cookie. The only API that supports cookies is AttachUprobeMulti, which uses `BPF_LINK_CREATE(BPF_TRACE_UPROBE_MULTI)`.

When a BPF program uses `bpf_get_attach_cookie()` to identify which function triggered an event, the only viable path for environments that do not support uprobe_multi is single-uprobe attachment with a per-uprobe cookie. Without this, `bpf_get_attach_cookie()` always returns 0, making per-function identification impossible.

### Proposal

Add a Go wrapper for `bpf_program__attach_uprobe_opts` that accepts at minimum a `bpf_cookie` parameter, returning a `*BPFLink` as with the existing `AttachUprobe`.

### References

* libbpf [`bpf_program__attach_uprobe_opts`](https://github.com/libbpf/libbpf/blob/f5dcbae736e5d7f83a35718e01be1a8e3010fa39/src/libbpf.h#L764)
* libbpfgo API tracking: https://github.com/aquasecurity/libbpfgo/issues/91

maxgio92:
Add a Go wrapper for `bpf_program__attach_uprobe_opts` that accepts at minimum a `bpf_cookie` parameter, returning a `*BPFLink` as with the existing `AttachUprobe`.

New selftests are included accordingly.

Closes #522

### DIFF
--- a/selftest/uprobe-cookie/main.go
+++ b/selftest/uprobe-cookie/main.go
@@ -1,0 +1,147 @@
+package main
+
+import "C"
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"hash/fnv"
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	bpf "github.com/aquasecurity/libbpfgo"
+	"github.com/aquasecurity/libbpfgo/selftest/common"
+)
+
+type Event struct {
+	Cookie uint64
+}
+
+type FunctionInfo struct {
+	Name   string
+	Offset uint64
+}
+
+const (
+	bpfProgramName   = "uprobe__test_functions"
+	bpfProgramObject = "main.bpf.o"
+)
+
+func main() {
+	if len(os.Args) < 3 {
+		common.Error(errors.New("wrong syntax"))
+	}
+
+	// Executable and expected symbols to be traced as positional arguments.
+	binaryPath := os.Args[1]
+	expectedSymbolNames := strings.Split(os.Args[2], ",")
+
+	// Build cookie-to-function mapping for expected symbols.
+	cookieToFunctionInfo := make(map[uint64]FunctionInfo)
+	type probeTarget struct {
+		offset uint64
+		cookie uint64
+	}
+	var targets []probeTarget
+
+	for _, name := range expectedSymbolNames {
+		offset, err := common.SymbolToOffset(binaryPath, name)
+		if err != nil {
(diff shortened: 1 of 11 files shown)


selftest/uprobe-cookie/main.go
```diff
@@ -0,0 +1,147 @@
...
+	// the expected functions.
+	time.Sleep(2 * time.Second)
+
+	// Verify that all uprobes have been executed.
+	for _, symbolName := range expectedSymbolNames {
+		if _, ok := got[symbolName]; !ok {
+			common.Error(fmt.Errorf("function %s has not been traced", symbolName))
+		}
+	}
+	log.Println("all functions have been traced with correct cookies")
```
maxgio92, line 132: The test is basically the same we do for uprobe multi as we mainly expect that all the uprobes are run relying on the cookies we got via the ring buffer.

selftest/uprobe-cookie/run.sh
```diff
@@ -0,0 +1,35 @@
...
+GOFUNCS="main.fooFunction,main.barFunction,main.bazFunction"
+
+# COMMON
+
+COMMON="$(dirname $0)/../common/common.sh"
+[[ -f $COMMON ]] && { . $COMMON; } || { error "no common"; exit 1; }
+
+# MAIN
+
+kern_version gt 5.15
```
maxgio92, line 20: because of `bpf_get_attach_cookie()`

geyslan:
@maxgio92 do you mind rebasing it to get the most recent tests in your branch?

geyslan:
github oddly didn't trigger the tests, so closing it in favour of https://github.com/aquasecurity/libbpfgo/pull/530

PR #530, opened 2026-06-24
feat: add support for bpf_program__attach_uprobe_opts

geyslan:
Close: #522

commit ac1ad5beba3c7f192221a880d03ecc2ffda61af3

    build(deps): lower root Go directive to 1.18
    
    Set the root go directive to 1.18 to match the repository's actual
    language and dependency floor.

commit 109ea40e6d9b4c5da9a655bad4b512e5c868f96d

    build: add Makefile go-get and go-tidy targets
    
    go-get and go-tidy are for maintaining nested modules.
    
    It also fixes selftests go.sum files by using go-tidy target.
    
    Additional fix was done to uprobe-multi module path.

commit 589730d44fd3a5f407e85195e2dcfa2fa328bc92

    test(selftest): add uprobe-cookie selftest
    
    Verify single-uprobe bpf_cookie attachment end-to-end. The BPF program
    reads the cookie via bpf_get_attach_cookie() and sends it back through
    a ringbuffer. The Go test attaches individual uprobes with FNV-hashed
    cookies to both C and Go target binaries and asserts all symbols are
    traced with the correct cookie values.
    
    Signed-off-by: Massimiliano Giovagnoli <maxgio92@pm.me>

commit dc76380b1f967d2c98ca505cd4bf915758c70d34

    feat(prog): add AttachUprobeWithOpts and AttachURetprobeWithOpts
    
    Expose single-uprobe attachment with bpf_cookie support in the Go API.
    AttachUprobeWithOpts and AttachURetprobeWithOpts wrap the new C helper
    and allow callers to set a per-attachment cookie retrievable in BPF
    programs via bpf_get_attach_cookie().
    
    Signed-off-by: Massimiliano Giovagnoli <maxgio92@pm.me>

commit 57e1d49ca2711f41285a672cbe8848561c183bf3

    feat: add C wrapper for bpf_program__attach_uprobe_opts
    
    Add cgo_bpf_program__attach_uprobe_opts() to support single-uprobe
    attachment with bpf_cookie, ref_ctr_offset, and retprobe options
    via bpf_uprobe_opts.
    
    Signed-off-by: Massimiliano Giovagnoli <maxgio92@pm.me>

PR #531, opened 2026-06-26
test(uprobe-cookie): cover uretprobe in selftest

maxgio92:
Adds uretprobe coverage to the `uprobe-cookie` selftest so it exercises `AttachURetprobeWithOpts`, not just `AttachUprobeWithOpts`.

PR #530 (and #523 before it) add both `AttachUprobeWithOpts` and `AttachURetprobeWithOpts`, but the selftest only attaches the entry uprobe. The uretprobe attach path has no coverage.

This PR is stacked on #530. This PR targets `main`, so its diff includes #530's commits until #530 merges. The uretprobe coverage itself is a single self-contained commit touching only the two selftest files.


geyslan:
Tks, updated the original with the new changes and fixed a lint error. Closing in favour of https://github.com/aquasecurity/libbpfgo/pull/530.

